PR #63, opened 2025-05-28
build: Remove pytest from runtime dependencies

### LINKED ISSUE #62: pytest is included as a required runtime dependency
```console
$ git grep "pytest" -- quantem/
```

shows no results, meaning that `pytest` is not used in any library code, however, it is included as a required runtime dependency

https://github.com/ophusgroup/quantem/blob/d37bca13379e8d2f1e3ac88b1f6f796077e593e4/pyproject.toml#L30-L42

Unused requirements should be avoided, and would be better suited for a `test` dependency group.

matthewfeickert:
Resolves #62 

**build: Remove pytest from runtime dependencies**

* Remove pytest from runtime dependencies as it is never used.
* Move pytest to a 'test' dependency group.

**chore: Update dependency group in uv lockfile**

**ci: Use test dependency group to install needed tools** 

* Don't install tools that aren't needed to run the tests, and install only the required tools from the 'test' dependency group.

### DIFF
--- a/.github/workflows/uv-pytests.yml
+++ b/.github/workflows/uv-pytests.yml
@@ -24,7 +24,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Install the project
-        run: uv sync --locked --all-extras --dev
+        run: uv sync --locked --all-extras --group test
 
       - name: Run pytests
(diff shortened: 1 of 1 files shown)


matthewfeickert:
This is ready for review.

Issue #61, opened 2025-05-28
PyPI distribution of quantem v0.1.4 does not package library correctly


matthewfeickert:
`quantem` `v0.1.4` on PyPI is broken at install as the library is not properly packaged.

## Minimal failing example

```console
(example) root@bc4314ce7113:/example# uv pip install quantem
Using Python 3.13.3 environment at: .pixi/envs/default
Resolved 40 packages in 1.96s
Prepared 40 packages in 32.29s
Installed 40 packages in 37ms
 + click==8.2.1
 + cloudpickle==3.1.1
 + cmasher==1.9.2
 + colorspacious==1.1.2
 + contourpy==1.3.2
 + crc32c==2.7.1
 + cycler==0.12.1
 + dask==2025.5.1
 + dill==0.4.0
 + donfig==0.8.1.post1
 + flexcache==0.3
 + flexparser==0.4
 + fonttools==4.58.0
 + fsspec==2025.5.1
 + h5py==3.13.0
 + iniconfig==2.1.0
 + kiwisolver==1.4.8
 + locket==1.0.0
 + matplotlib==3.10.3
 + numcodecs==0.16.1
 + numpy==2.2.6
 + packaging==25.0
 + partd==1.4.2
 + pillow==11.2.1
 + pint==0.24.4
 + platformdirs==4.3.8
 + pluggy==1.6.0
 + pyparsing==3.2.3
 + pytest==8.3.5
 + python-box==7.3.2
 + python-dateutil==2.9.0.post0
 + pyyaml==6.0.2
 + quantem==0.1.4
 + rosettasciio==0.8.0
 + scipy==1.15.3
 + six==1.17.0
 + toolz==1.0.0
 + tqdm==4.67.1
 + typing-extensions==4.13.2
 + zarr==3.0.8
(example) root@bc4314ce7113:/example# python
Python 3.13.3 | packaged by conda-forge | (main, Apr 14 2025, 20:44:03) [GCC 13.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import quantem
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import quantem
  File "/example/.pixi/envs/default/lib/python3.13/site-packages/quantem/__init__.py", line 1, in <module>
    from quantem.core import io as io
ModuleNotFoundError: No module named 'quantem.core'
>>> 
```

## Diagnosis 

[...]

PR #65, opened 2025-05-29
automatically tagging release branches

gvarnavi:
Addresses @matthewfeickert aside in #61 by automatically tagging releases.


gvarnavi:
Right. I suppose tags is fine for now and we can switch to releases once we figured out how to release notes and the like?

matthewfeickert:
> I suppose tags is fine for now and we can switch to releases once we figured out how to release notes and the like?

Sure. GitHub releases will autoprompt you to generate release notes from the commits since the last release, though some of that might not be "user focused" and you might want to curate it a bit.

PR #78, opened 2025-05-30
fix: Remove giving subdir path as package

matthewfeickert:
Resolves https://github.com/ophusgroup/quantem/issues/61 (c.f. https://github.com/ophusgroup/quantem/issues/61#issuecomment-2921327542)

Amends PR https://github.com/ophusgroup/quantem/pull/64

This PR seems to indicate that the tests aren't working properly, and so aren't properly installing the library from development.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,6 +37,15 @@ dependencies = [
   "h5py>=3.13.0",
 ]
 
+[tool.hatch.build.targets.sdist]
+# hatchling always includes:
+# pyproject.toml, .gitignore, any README, any LICENSE, AUTHORS
+only-include = [
+    "/src",
+    "/tests",
+    "/CITATION.cff"
+]
+
 [dependency-groups]
 test = [
     "pytest>=8.3.5",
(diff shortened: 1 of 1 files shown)


gvarnavi (approved):
Oh I see, it's to prevent additional non-necessary files per your comment. Thanks -- approved and merged!

matthewfeickert:
> Any reason for the explicit?
> 
> ```toml
> [tool.hatch.build.targets.sdist]
> only-include = [
>     "/src",
>     "/tests",
>     "/CITATION.cff"
> ]
> ```

Ah yeah, I was lazy and didn't summarize the commits in the PR body. Commit https://github.com/ophusgroup/quantem/pull/78/commits/624c6c0d7f08ce0d9647138b2f3f1adf250042c5 message summarizes 

> By default hatch will package the entire repository in an sdist for reproducible builds. It is standard to only distribute the source code and tests though for repackaging systems and to decrease storage.

Without this you're distributing _everything_ in your repository in your sdist and there's probably stuff that people don't really care about like GitHub Actions and docs. So it just simplifies things and makes the sdist smaller / decreases PyPI bandwidth.

PR #93, opened 2025-06-06
Diffractive imaging into dev --  torch ptycho 

arthurmccray:
Includes core/ml files, utils for array/tensor manipulation, and torch 
Went from dev -> tomography -> diffractive imaging recently so this shouldn't be an issue.


src/quantem/core/quantem.yaml
```diff
@@ -18,14 +21,48 @@ mkl:
...
   suppress-all-: false
-visualize:
+viz:
+  # default interpolation method to use for plotting
+  # "nearest" is the default of imshow, "none" is no interpolation and the default of matshow
+  interpolation: "nearest"
   # The default units to use in real space
   real_space_units: "Angstrom"
   # The default units to use in reciprocal space
   reciprocal_space_units: "Angstrom"
```
cophus, line 31: Shouldn't `reciprocal_space_units` be 1/Angstrom or Angstrom^-1?

cophus (commented):
Few minor changes, otherwise looks good!

PR #96, opened 2025-06-18
Tomography Module: SIRT Reconstruction, and Voxelwise AD

### LINKED ISSUE #80: Implementing AD Tomography Reconstruction
- [ ] Add projection operator
- [ ] Build ML in `core`

### LINKED ISSUE #82: Background/Intensity Subtraction for Tomo Series
- [ ] Still can't get perfect background in SIRT reconstruction

### LINKED ISSUE #83: Moving tilt series dataset to `Dataset3D`
Current implementation of `TiltSeries` is not how `Dataset3D` was intended to be used. Need to either implement the tilt angle into `origin` or `sampling`.

cedriclim1:
~~Test notebook + data: [Tomography_quantem.zip](https://github.com/user-attachments/files/20804377/Tomography_quantem.zip)~~
Updated Notebook (July 6, 2025):  [Tomography_Tutorial.zip](https://github.com/user-attachments/files/21091760/Tomography_Tutorial.zip)

The current iteration of the Tomography module's reconstruction algorithms are fully `torch` based:

1. SIRT reconstruction - A port of `scikit_image`'s implementation of `radon` and `iradon` to `torch`.
2. Voxelwise AD reconstruction - Automatic-differentiation loop directly optimizing the volume.

All of this is called from a top-level class `Tomography` (in `tomography.py`) which inherits from `TomographyConv` (`tomography_conv.py`), `TomographyML` (`tomography_ml.py`), and `TomographyBase` (`tomography_base.py`).

- `TomographyConv` inherits from `TomographyBase` and contains the conventional reconstruction algorithms (currently just SIRT). 
- `TomographyML` is almost a direct copy of `PtychographyML` except the optimizable values are the volume, tilt angles, and shifts.
- `TomographyBase` inherits from `AutoSerialize` and has the method `from_tilt_series` to instantiate the class. The tilt series dataset here is a `TomographyDataset` class which has fully `torch` variables. Since all reconstruction algorithms use `torch`, this is the easiest way to implement this without switching back from `numpy` to `torch` constantly. 

Following the design of `diffractive_imaging` I also have implemented a `object_models.py` file which contains the object (the volume). In the class, I can apply soft and hard constraints upon calling `.forward` on the object. 

In `utils.py` there are also a few utility functions for preprocessing raw data, prior to loading data into `Tomography`.

Also added `scikit_image`, `tensorboard`, and `torchvision` as additional dependencies.


cophus:
A few import issues:

```
/home/cophus/repos/quantem/.venv/lib/python3.12/site-packages/tqdm/auto.py:21: TqdmWarning: IProgress not found. Please update jupyter and ipywidgets. See https://ipywidgets.readthedocs.io/en/stable/user_install.html
  from .autonotebook import tqdm as notebook_tqdm
/home/cophus/repos/quantem/.venv/lib/python3.12/site-packages/colorspacious/comparison.py:11: SyntaxWarning: invalid escape sequence '\D'
  """Computes the :math:`\Delta E` distance between pairs of colors.
```

src/quantem/tomography/tomography_base.py
```diff
@@ -34,132 +47,132 @@ def __init__(
...
         self._loss = []
         self._mode = []
 
+        self._hard_constraints = self.DEFAULT_HARD_CONSTRAINTS.copy()
+        self._soft_constraints = self.DEFAULT_SOFT_CONSTRAINTS.copy()
+
+        self._logger = logger
+
     @classmethod
     def from_tilt_series(
```
cophus, line 67: I think "from_tilt_series" is a bit confusing relative to "from_data".  Seeing as how you made a tilt_series class, I would suggest these inits:
.from_tilt_series  - from TiltSeries class
.from_array          - from numpy arrays
.from_file              - from a single file, or perhaps also a list of files (if for example each tilt angle is a separate file)

src/quantem/tomography/tomography_base.py
```diff
@@ -306,6 +387,9 @@ def plot_projections(
         loss: bool = False,
         fft: bool = False,
     ):
+        volume_obj_np = self.volume_obj.obj.detach().cpu().numpy()
```
cophus, line 390: plot_projections is way overloaded.  You should separate these out:

.plot_loss()    (which could be a flag in .sirt_recon() which is set to True by default.)
.plot_projections()
.plot_projections(fft=True)

Then the user can generate just the plots they want.  I also suggest making .plot_loss() etc. have their own default figsize (e.g. for loss I'd use something like figsize=(8,4)) and also an input figax. This way the user can build whatever layout they want.

src/quantem/tomography/tomography_base.py
```diff
@@ -316,19 +400,19 @@ def plot_projections(
             fig, ax = plt.subplots(ncols=3, figsize=(20, 8))
 
         show_2d(
-            self.recon_volume.array.sum(axis=0),
+            volume_obj_np.sum(axis=0),
```
cophus, line 403: Doesn't summing along axis = 0 create a y-z projection?

src/quantem/tomography/tomography_base.py
```diff
@@ -316,19 +400,19 @@ def plot_projections(
...
             figax=(fig, ax[0]),
             cmap=cmap,
             title="Z-X Projection",
         )
         show_2d(
-            self.recon_volume.array.sum(axis=1),
+            volume_obj_np.sum(axis=1),
```
cophus, line 409: Doesn't summing along axis = 1 create a x-z projection?

src/quantem/tomography/tomography_base.py
```diff
@@ -316,19 +400,19 @@ def plot_projections(
...
             figax=(fig, ax[1]),
             cmap=cmap,
             title="Y-X Projection",
         )
         show_2d(
-            self.recon_volume.array.sum(axis=2),
+            volume_obj_np.sum(axis=2),
```
cophus, line 415: Same comment here:
Doesn't summing along axis = 2 create a x-y projection?

The x-y-z convention (and how it maps onto dims) you choose should be clearly outlined in the docstring.

src/quantem/tomography/tomography_base.py
```diff
@@ -338,20 +422,20 @@ def plot_projections(
             fig, ax = plt.subplots(ncols=3, figsize=(25, 8))
 
             show_2d(
-                np.abs(np.log(np.fft.fftshift(np.fft.fftn(self.recon_volume.array.sum(axis=0))))),
```
cophus, line 341: I strongly dislike default log scale plotting!  This is why show_2d supports log and power law scaling.  If you make this function:
.plot_projections(
    ...
    fourier_transform=True,
    **kwargs,
)
you can pass **kwargs through to show (with defaults you want to specify, removed with "pop").

src/quantem/core/utils/utils.py
```diff
@@ -3,7 +3,7 @@
 from typing import TYPE_CHECKING, Any, Iterator, List, Optional, Sequence, Tuple
 
 import numpy as np
-from tqdm import tqdm
+from tqdm.auto import tqdm
```
cophus, line 6: I didn't know about this auto import, neat!

src/quantem/tomography/radon/radon.py
```diff
@@ -0,0 +1,212 @@
+"""
+A fully torch based implementation of the Radon transform based on compatible with CPU and GPU.
```
cophus, line 2: sentence fragment

src/quantem/tomography/tilt_series_dataset.py
```diff
@@ -18,6 +19,9 @@ def __init__(
         sampling: NDArray | tuple | list | float | int,
         units: list[str] | tuple | list,
         tilt_angles: list | NDArray,
+        z1_angles: list | NDArray,
```
cophus, line 22: The docstring should fully explain the naming convention used here for angles, what shifts are, the units for angles and shifts, the directions (e.g. is the image shifted with respect to the original data?)

src/quantem/tomography/tomography_base.py
```diff
@@ -34,132 +47,132 @@ def __init__(
             A descriptive name for the dataset
         """
 
-        if _token is not self._token:
-            raise RuntimeError(
-                "This class is not meant to be instantiated directly. Use the from_data method."
-            )
+        # if _token is not self._token:
+        #     raise RuntimeError(
+        #         "This class is not meant to be instantiated directly. Use the from_data method."
```
cophus, line 52: Or the .from_tilt_series() method right?  data is ambiguous (could be TiltSeries, numpy, or files/list of files)

src/quantem/tomography/tomography_base.py
```diff
@@ -34,132 +47,132 @@ def __init__(
...
+        #     )
 
         self._device = device
-        self._tilt_series = tilt_series
-        self._recon_volume = recon_volume
+        self._dataset = dataset
+        self._volume_obj = volume_obj
```
cophus, line 57: Should the reconstructed volume be a private variable?  Or do you have other views defined later?

src/quantem/tomography/tomography_dataset.py
```diff
@@ -0,0 +1,212 @@
...
+
+from quantem.core.datastructures.dataset3d import Dataset3d
+from quantem.core.io.serialize import AutoSerialize
+from quantem.core.utils.validators import (
+    validate_array,
+    validate_tensor,
+)
+
+
+class TomographyDataset(AutoSerialize):
```
cophus, line 13: What is the point of having TomographyDataset and TiltSeries?  It seems like these 2 classes do the same thing.

src/quantem/tomography/tomography_logger.py
```diff
@@ -0,0 +1,70 @@
+import matplotlib.pyplot as plt
+
+from quantem.core.ml.logger import LoggerBase
+from quantem.tomography.object_models import ObjectModelType
+from quantem.tomography.tomography_dataset import TomographyDataset
+
+
+class TomoLogger(LoggerBase):
```
cophus, line 8: I'd suggest naming conventions like:
LoggerTomography

I prefer the parent name coming first if possible to make finding things easier when looking through attributes.

src/quantem/tomography/tilt_series_dataset.py
cedriclim1: See: https://github.com/ophusgroup/quantem/pull/96#discussion_r2188389147

src/quantem/tomography/tomography_base.py
cedriclim1: See: https://github.com/ophusgroup/quantem/pull/96#discussion_r2188389147

src/quantem/tomography/tomography_base.py
cedriclim1: In Line 151:
```
    @property
    def volume_obj(self) -> Dataset3d | ObjectModelType | None:
        """Reconstruction volume dataset."""

        return self._volume_obj
```

src/quantem/tomography/tomography_base.py
cedriclim1: OK, I admit, the current volume convention is (z, y, x) which is a little psychotic. This can be changed, but I think as long as we're explicit saying that this is the axes we're working in it should be ok...?

src/quantem/tomography/tomography_base.py
cedriclim1: See: https://github.com/ophusgroup/quantem/pull/96#discussion_r2188413530

PR #101, opened 2025-09-18
adding metadata dictionary

smribet:
This PR is adding metadata as dictionary to Dataset and Dataset4dstem. This could be converted to a metadata class down the line.


cophus (approved):
Looks good!

PR #106, opened 2025-09-28
Optimized Center of Mass Functionality

henryhng:
Add an optimized center of mass routine for diffraction patterns with support for CPU & Apple MPS.
comprehensive_performance_results


henryhng:
**October 5, 2025**

- Moved COM implementation into a new diffraction DPC package
- Skipped caching on GPU to avoid VRAM leaks
- Removed SciPy single-pattern shortcut so all simple cases (single and batch) use vectorised Torch path.

20251005-com_updates

gvarnavi:
Cool to see `torch.compile` use-cases! Curious on the benchmark graphs: JIT compiling indeed achieves a 300x speed-up over a vectorized call of the sort `torch.sum(intensities * qxa[None, :, :], dim=(-2, -1)) / summed_intensities` ??

A quick note on abstraction / organization. I think a fast center_of_mass function should live in utils, and then called in appropriate OriginModels (which need to calculate CoM, fit the origin, shift etc). [Here's one such implementation](https://github.com/electronmicroscopy/quantem/blob/ae3294f1a399c6057cf167e90b6e19b9126a3c50/src/quantem/diffractive_imaging/origin_models.py#L16) I started forever ago, hoping to have time for quantem dev later this month..

PR #114, opened 2025-10-14
Added SIREN, HSIREN, and Finer with allowed complex inputs

cedriclim1:
Added SIREN, HSIREN, and Finer models.


src/quantem/core/ml/activation_functions.py
arthurmccray: looks good

src/quantem/core/ml/blocks.py
arthurmccray: not going through the actual sine blocks and such, trusting it works.

PR #125, opened 2025-11-19
merging diffractive_imaging into dev to coincide with DGP paper

arthurmccray:
This PR brings some new features, bugfixes, and general improvements. There are a couple of changes outside `diffractive_imaging`, with the only semi-significant one being a bugfix and simplification in `autoserialize`. 

Example notebooks can be found at [quantem-tutorials/ml](https://github.com/electronmicroscopy/quantem-tutorials/tree/ml) 

Still to do -- create an example notebook showing the direct ptycho -> iterative ptycho workflow (and potentially smooth that over with some new methods)

### DIFF
--- a/src/quantem/core/datastructures/dataset.py
+++ b/src/quantem/core/datastructures/dataset.py
@@ -645,7 +645,7 @@ class Dataset(AutoSerialize):
                 cls = Dataset
 
         # Construct new dataset
-        return cls.from_array(
+        return cls.from_array(  # type: ignore ## would be nice to properly type slicing, but hard
             array=array_view,
             name=f"{self.name}{index}",
             origin=new_origin,


src/quantem/core/io/serialize.py
arthurmccray: Previously I was having issues where autoreload was causing autoserializable subclasses to not be recognized (this is due to `isinstance` failing with autoreload, which seems to be a known issue). I had fixed it, but somehow that caused a bug where optimizer states were not being re-loaded about 50% of the time. I have no idea what caused this, but rerunning the same cell of a notebook repeatedly did not give the same results every time. 

This is (I think) a fix to both issues. I've tested it and it seems pretty robust so far, but time will tell...

src/quantem/diffractive_imaging/logger_ptychography.py
arthurmccray: epoch -> iter

src/quantem/core/ml/optimizer_mixin.py
```diff
@@ -106,6 +113,7 @@ def set_optimizer(self, opt_params: dict | None = None) -> None:
 
         if isinstance(opt_type, type):
             self._optimizer = opt_type(params, **opt_params)
+            self.optimizer_params["type"] = opt_type.__module__.split(".")[-1]
```
gvarnavi: this seems dangerous / hacky.. is there a more robust way to achieve what you're trying to do? where is this used?

src/quantem/core/utils/diffractive_imaging_utils.py
gvarnavi: These should probably move to a more general utils file, like we need to fit a probe circle for all non-calibrated datasets

src/quantem/diffractive_imaging/dataset_models.py
```diff
@@ -63,6 +64,8 @@ def __init__(
         self.verbose = verbose
         self._preprocessed = False
         self._preprocessing_params = {}  # for serialization and reloading
+        self._com_rotation_rad = 0  # default
```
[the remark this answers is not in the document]
gvarnavi: I think Steph wants us to switch to specifying everything in degrees, but ok for now to be consistent with direct ptycho

src/quantem/diffractive_imaging/dataset_models.py
```diff
@@ -164,6 +169,23 @@ def positions_px_fractional(self) -> torch.Tensor:
     # endregion --- optimizable parameters ---
 
     # region --- buffers ---
+    @property
```
gvarnavi: I suspect this is used elsewhere to mask updates?

src/quantem/diffractive_imaging/ptychography_base.py
```diff
@@ -476,12 +520,15 @@ def probe_model(self, model: ProbeModelType | type):
...
-        )
+        try:
+            self._probe_model.set_initial_probe(
+                self.roi_shape,
+                self.reciprocal_sampling,
+                self.dset.mean_diffraction_intensity,
+                device=self.device,
+            )
+        except Exception:
+            pass
```
gvarnavi: why is this allowed to fail silently again?

src/quantem/diffractive_imaging/ptycho_utils.py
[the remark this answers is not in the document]
arthurmccray: I added a `val_ratio` attribute (intended to be set during `ptycho.preprocess`) that allows for designating some fraction of the scan positions as validation data, so the loss will be recorded but they will not be used to train. This is potentially useful for visualizing overfitting to noise, though in practice you'll of course want to run your final recons learning from 100% of the scan positions. 

And to confirm, nothing is changed when `val_ratio` is left at the default value of `0`.

src/quantem/diffractive_imaging/ptychography_base.py
arthurmccray: This only fails if the dset hasn't been preprocessed, and is then called in ptycho.preprocess instead. I'm updating it to not be an explicit if/else call rather than try/except with a comment to explain.

src/quantem/core/datastructures/dataset4d.py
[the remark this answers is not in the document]
arthurmccray: Seems reasonable to me. I don't think there's a way to make it play nicely with static type checking, at least not easily.

gvarnavi (approved):

PR #131, opened 2026-01-06
updating core/ml models

arthurmccray:
This PR expands on the neural networks included in /core/ml, and generally cleans up that code with docstrings, linter fixes, etc. This is also associated with updated tutorial notebooks in [draft PR at quantem-tutorials](https://github.com/electronmicroscopy/quantem-tutorials/pull/3) and which will be pulled into main there when done. 

In general this adds some new models and refactors a few things, but I think everything should backwards-compatible and non-breaking. I also removed the `Finer` model because I think we are wanting folks to use `HSiren` for INRs in general.


gvarnavi:
Heads-up: you'll need to pull dev into the PR to enable the updated automated checks.

pyproject.toml
cedriclim1, line 1: Assuming everyone is happy having torchinfo as a dependency?

src/quantem/core/config.py
cedriclim1, line 1: Change `str` as a Literal.

src/quantem/core/ml/dense_nn.py
cedriclim1, line 1: I don't like calling layers `blocks`, I feel like this terminology works with U-net's I think.

cedriclim1 (commented):
Left some comments

pyproject.toml
arthurmccray: yeah, `torchinfo` seems to be actively maintained, and it doesn't add any additional dependencies itself.

src/quantem/core/config.py
arthurmccray: ended up leaving it as a `str` to account for passing `torch.device`-like strings, e.g. `"cuda:2"`

src/quantem/core/ml/dense_nn.py
arthurmccray: sure, changed to `layer`

cedriclim1 (approved):

src/quantem/core/config.py
```diff
@@ -503,33 +503,35 @@ def validate_device(dev: str | int | torch.device | None = None) -> tuple[str, i
         dev = torch.device(
             "cuda" if torch.cuda.is_available() else "mps" if torch.mps.is_available() else "cpu"
         )
-    elif isinstance(dev, str) and dev.lower() == "gpu":
```
gvarnavi, line 506: @arthurmccray @cedriclim1 Why did we drop `device='gpu'` support?

arthurmccray: ah shoot, that's a mistake! Didn't realize that wasn't in the tutorial notebook that i had used for testing.  I'll make a hotfix branch and fix that along with adding a proper pytest so it can't happen again.

PR #119, opened 2025-10-27
In situ time series alignment

ceobrero:
## What this PR does...

Takes an unaligned time series image stack and applies 2D cross-correlation to correct for drift.

Includes options to pad and edge blend images.
### Example of applying padding and edge blending to an unaligned time series image stack series:

`padded = series.pad_and_blend(pad_width = 16, edge_blend = 16, pad_val = 0)
`
511427432-f164b709-99bd-49b9-8439-d49914c8c3bc

Using 2D cross correlation on an image or multiple images (image stack)
### Example of applying 2D cross-correlation on an image (final frame) to a reference image (initial frame) from a padded and edge-blended time series image stack padded:

`shift, im__shift = em.imaging.time_series.TimeSeries.align_image(padded[0], padded[-1])
`

Showing the combined image + ref image before and after alignment:
511474807-5d8dfb54-df5a-466b-8222-d41553eac23c

### Example of applying 2D cross-correlation on a padded and edge-blended time series image stack padded:

`aligned, dxy = em.imaging.time_series.TimeSeries.align_stack(padded)
`

Showing the image stack mean before and after alignment:
511429420-7e747848-696c-4b03-9e41-6b731d59444e

Example Notebook:
[dev_time_series_example (4).ipynb](https://github.com/user-attachments/files/24601729/dev_time_series_example.4.ipynb)

PR #147, opened 2026-01-13
In situ time series alignment class

ceobrero:
## What this PR does...

Note: updated from PR #119. TimeSeries class is now object-oriented and serialized. Pytests included.

Takes an unaligned time series image stack and applies 2D cross-correlation to correct for drift. The image stack is preprocessed (padded and edge-blended) upon instantiation. The subpixel shifts are determined on an image stack that is filtered via 2D convolution to accentuate features (edges) to align with.

Example of applying 2D cross-correlation on a preprocessed time series image stack:

`binned_pad.align_stack(running_average_frames = 10, sigma_edge = 0.7, sf_val = 0.5)`

Visualizing the image stack mean before and after alignment:
unalignedvsaligned

Example Notebook: 
[dev_time_series_example.ipynb](https://github.com/user-attachments/files/24597567/dev_time_series_example.ipynb)


ceobrero:
@bobleesj updated this time series alignment PR, ready for review!

src/quantem/core/utils/imaging_utils.py
```diff
@@ -548,6 +549,61 @@ def fourier_cropping(
...
+        2D array containing the gradient magnitude (edge strength) of the input image.
+    """
+
+    sf_val = np.array(sf_val)
+    if sf_val.size == 1:
+        sf_val = np.array(
+            [sf_val, sf_val],
+            )
+
+    r = np.arange(
```
bobleesj: `sf_val` is a bit hard to guess what this parameter means without looking at the API. the intenral variable names  can be also improved.

src/quantem/core/utils/imaging_utils.py
```diff
@@ -548,6 +549,61 @@ def fourier_cropping(
...
+        -np.ceil(2.0*sigma_edge),
+        np.ceil(2.0*sigma_edge+1),
+    )
+
+    k = np.exp(
+        (r[:,None]**2) / (-2*sigma_edge**2)
+    )
+
+    sf = np.array([
```
bobleesj: also, if we end-up using a simple variable name, we want to write as compcat as possible like writing a series of equation, etc without uncessasry blank lines (please use the below as example only)

```
def _gaussian_1d_kernel(sigma: float, device, dtype):
    # Match your r range: [-ceil(2*sigma), ..., ceil(2*sigma)]
    radius = int(torch.ceil(torch.tensor(2.0 * sigma)).item())
    r = torch.arange(-radius, radius + 1, device=device, dtype=dtype)
    k = torch.exp(-(r * r) / (2.0 * sigma * sigma))
    k = k / k.sum()  # normalize (optional but usually desirable)
    return k  # shape (K,)
```

bobleesj (commented):
Thanks @ceobrero


General comment on performance since for times-series, performance is the primary concern (just like in DriftCorrection).

-  We want to do FFT at once across all images with vectorized operations instead of applying FFT across each loop.
sth like:
```
 # FFT entire stack at once, multiply, IFFT back
      F = np.fft.fft2(images, axes=(-2, -1))
      F_filtered = F * edge_kernel  # Broadcasting: (N,H,W) * (H,W)
      filtered = np.abs(np.fft.ifft2(F_filtered, axes=(-2, -1)))
```

- For filtering, etc. we want to utilize fourier if possible. Right now I see there are 6 convolutions are being called but we can do this in a single pass. 

Sth like:
```
      kx = np.fft.fftfreq(H)[:, None]
      ky = np.fft.fftfreq(W)[None, :]
      gaussian = np.exp(-2 * (np.pi * sigma_edge)**2 * (kx**2 + ky**2))
      gradient_mag = np.sqrt((2 * np.pi * kx)**2 + (2 * np.pi * ky)**2)
      edge_kernel = gaussian * gradient_mag
```

- Since it's time-series and it must be done sequentially w/ moving average, we probalby want to have some sort of progress:

```
from tqdm import tqdm
  for a0 in tqdm(range(nframes - 1), desc="Aligning"):
```

- Parabolic fitting vs.. DFT upsampling (already in cross_correlation_shift), which one do we choose?

Also, if you could share, a performance chart would be good. So that we know how long it would take process a batch of images across size dimensions.

src/quantem/core/utils/imaging_utils.py
ceobrero: `sf_val` is changed to 'sobel_filter_val', also changed docstring to reflect the variable better.

PR #150, opened 2026-01-17
`Dataset3d` support `show()` 

bobleesj:
### What this PR does

Visualize time-series, tomography, etc.

Screenshot 2026-01-23 at 7 57 41 AM

Screenshot 2026-01-23 at 7 57 52 AM

Screenshot 2026-01-23 at 7 57 59 AM

Screenshot 2026-01-23 at 7 58 12 AM

API design

```python
data.show()                          # first 20 frames (safe default)
data.show(max=None)                  # all frames
data.show(start=5, max=1)            # single frame
data.show(start=-1, max=1)           # last frame (negative indexing)
data.show(start=10, end=20)          # frames 10-19
data.show(step=5)                    # every 5th frame
data.show(start=9, step=-1)          # reverse order
data.show(ncols=2)                   # 2 columns
data.show(title_prefix="DP")         # "DP 0", "DP 1", ...
data.show(suptitle="Time Series")    # figure title
data.show(cmap="viridis", cbar=True) # colormap with colorbar
data.show(scalebar=True)             # enable scalebar
fig, axes = data.show(returnfig=True)# get figure for customization
```

Physical notebook:

[20260123_dataset3d_visual_demo_v2.ipynb](https://github.com/user-attachments/files/24824503/20260123_dataset3d_visual_demo_v2.ipynb)

Online notebook:

https://colab.research.google.com/github/bobleesj/quantem-examples/blob/main/dataset3d_visualization.ipynb

### What should PR reviewer(s) do

Please feel free to test the notebook and evaluate its API design

### DIFF
--- a/src/quantem/core/datastructures/dataset3d.py
+++ b/src/quantem/core/datastructures/dataset3d.py
@@ -221,7 +221,7 @@ class Dataset3d(Dataset):
         returnfig : bool, default False
             If True, returns (fig, axes).
         **kwargs : dict
-            Keyword arguments for show_2d (cmap, cbar, norm, etc.).
+            Keyword arguments for show_2d (cmap, cbar, vmin, vmax, norm, etc.).
 
         Returns
         -------
@@ -287,15 +287,26 @@ class Dataset3d(Dataset):
                 f"Start {start} is out of bounds for dataset with {total_frames} frames. "
                 f"Please use start in range [0, {total_frames - 1}]."
             )
-        # Compute frame indices (avoid creating huge intermediate list)
-        end_idx = end if end is not None else (total_frames if step > 0 else -1)
+
+        # Compute frame indices
+        # Default end: go forward to total_frames, or backward to -1 (before index 0)
+        if end is not None:
+            end_idx = end
+        else:
+            end_idx = total_frames if step > 0 else -1
+
+        # Clamp to valid range (only needed for positive step)
         if step > 0:
             end_idx = min(end_idx, total_frames)
-            if max is not None:
-                end_idx = min(end_idx, start + max * step)
-        else:
-            if max is not None:
-                end_idx = max(end_idx, start + max * step)
+
+        # Apply max limit to avoid creating huge list
+        if max is not None:
+            max_end = start + max * step
+            if step > 0:
+                end_idx = min(end_idx, max_end)
+            elif max_end > end_idx:
+                end_idx = max_end
+
         frame_idx = list(range(start, end_idx, step))
         if len(frame_idx) == 0:
             raise ValueError(
@@ -311,7 +322,8 @@ class Dataset3d(Dataset):
             for i in frame_idx
         ]
         # Pad last row to complete the grid (show_2d requires rectangular input)
-        pad_count = (-n_frames) % ncols
+        remainder = n_frames % ncols
+        pad_count = 0 if remainder == 0 else ncols - remainder
         if pad_count > 0:
             images.extend([np.zeros_like(self.array[0])] * pad_count)
             labels.extend([""] * pad_count)
--- a/src/quantem/core/datastructures/dataset3d.py
+++ b/src/quantem/core/datastructures/dataset3d.py
@@ -287,35 +287,45 @@ class Dataset3d(Dataset):
                 f"Start {start} is out of bounds for dataset with {total_frames} frames. "
                 f"Please use start in range [0, {total_frames - 1}]."
             )
-
-        # Compute frame indices
+        # Compute frame indices (avoid creating huge intermediate list)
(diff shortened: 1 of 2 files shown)


bobleesj:
@ceobrero Could you pls review this and stress test this?  since you have thousands of times-series data, that would quite help!

ceobrero:
@bobleesj I tested this out using one of my +1000 frame time-series datasets and ran into a couple of issues. After instantiating the dataset, running ().show on all frames was very slow and took up a lot of memory on my PC (subsequently causing it to crash).

image

It would be great if there was some way to select a range of frames to show, show every Nth frame, show a max number of frames, etc. I also think, if not now then in the future, it would be good to have the option to show a time-series slider to scroll through multiple frames; showing frames as a grid is fine for a small subset of frames, but having a slider would be much more convenient for time-series data. 

Unsure if this is out of the scope of this PR, but is there a way to show volumetric data (for tomography)?

Since both time-series and tomography data fall under the umbrella of 3D datasets, do you think there should be some differentiation between showing the two? I would say "slice" is more of a tomography term, while "frame" is more of a time-series term.

bobleesj:
@ceobrero Thanks for testing. That's a pretty solid stress test. How did it go with 100 images, something reasonable? It's hard to tell whether the memory thing is based on the inherent matplotlib API.

Since `dataset3d` natively supports slicing like `dset[3:10]`, I didn't duplicate that feature as a function but based on the below comment of yours, I can probably add that within `show()`

> show every Nth frame, show a max number of frames,

yup, will do. 

> In the future, it would be good to have the option to show a time-series slider 

That works. Pls create an GH issue? Let's see  if I can make a widget later since I'm working with time-series. \

FYI - slider (anything interactive) isn't rendered as a preview when we upload jupyter notebooks to GitHub so we want to separate widget vs. matplotlib's show in general.

> Unsure if this is out of the scope of this PR, but is there a way to show volumetric data (for tomography)?

This function just shows images based on the datatype present in `dataset3d` so idk how to extend this atm. So, probably another API design needed or a new GIthub for another discussion.

> do you think there should be some differentiation between showing the two? I would say "slice" is more of a tomography term, 

Good point - `frame` is more agnostic than `slice` since we are working in the context of image (show2d), meaning there is a frame for an each image. We want `dataset3d` API names to be general while intuitive so I'm not sure. probably for tomography, one may want to subclass `dataset3d` instead in the future - just like `dataset4dstem` Unless you have a strong alternative for a general api name/internal variable name within `dataset3d`, pls let me know if `frame` is ok.

Things to follow-up

- [x] @ceobrero  Stress test - How about 100 images?
- [ ] @ceobrero Widget slider for time-series - anywidget
- [x] Support nth indexing for plotting
- [x] Support start_frame, end_frame within` show`

arthurmccray:
i'm not sure where things stand with the widget stuff, but I think it would be fine to merge if you we add a version of `max_images` and `every_nth`. The main thing is the `max_images` with a sensible default (20?), we don't want someone calling `.show()` for a large dset to crash the kernel or be unnecessarily slow. Widget/interactive stuff can be in a different PR. 

I get the logic of already supporting slicing in `dset`, such that `every_nth` is redundant with `dset[::n]`, but I would prefer to still include that in the `show` function as the frame titles will then be correct. 

For anything more complicated than that, e.g. `start_frame` and `end_frame`, I think it would probably just be easier to allow users to pass an `indices: list[int] | None` which would take priority, and allow for whatever specifics the user wants.  

Also I agree with this basic `show` function being datatype agnostic, i.e. the same for volumetric and time series data. Subclasses can overload it for specifics like titles.

ceobrero:
@bobleesj 
>  @/ceobrero Stress test - How about 100 images?

Just tested on 100 images and 500 images; 100 works pretty well (~10s on my PC), and I would say anything >400 frames would be a no-go. I agree with @arthurmccray about setting a default, maybe default `max_frames` is 20, but then the maximum is 100 or 200? 

> Unless you have a strong alternative for a general api name/internal variable name within dataset3d, pls let me know if frame is ok.

> Also I agree with this basic show function being datatype agnostic, i.e. the same for volumetric and time series data. Subclasses can overload it for specifics like titles.

I think `frame` is a good general variable name!

bobleesj:
@ceobrero @arthurmccray plese test - notebook updated in the PR comment above

  - [x] slice-like API: start, end, step, max                                                                                               
  - [x] safe default: max=20 prevents memory issues                                                                                         
  - [x] negative indexing and negative step support                                                                                         
  - [x] input validation with helpful error messages                                                                                        
  - [x] absolute suptitle spacing (consistent across row counts)

bobleesj:
test added

arthurmccray (approved):
lgtm! @ceobrero can you give this a quick test with one of your datasets please?

PR #153, opened 2026-01-20
Hotfix config

arthurmccray:
Adding back `"gpu"` as an option for config setting device after it was accidentally removed,  and updating the tests to prevent it happening in the future. 

I also added an attribute to some of the classes in `test_autoserialize.py` as it was raising warnings. The issue was the `TestingClass` and others which were meant as classes to be used in tests, but which were trying to be ran as tests because of the name. Fixed by adding a `"__test__" = False` line as suggested [here](https://stackoverflow.com/questions/62460557/cannot-collect-test-class-testmain-because-it-has-a-init-constructor-from).

### DIFF
--- a/src/quantem/core/config.py
+++ b/src/quantem/core/config.py
@@ -506,6 +506,8 @@ def validate_device(dev: str | int | torch.device | None = None) -> tuple[str, i
     elif isinstance(dev, str):
         if "cuda" in dev.lower():
             dev = torch.device(dev)
+        elif "gpu" in dev.lower():
+            dev = torch.device("cuda")
         elif dev.lower() == "mps":
             dev = torch.device("mps")
         elif dev.lower() == "cpu":
@@ -572,6 +574,7 @@ def set_device(dev: str | int | "torch.device") -> None:
     >>> set_device(torch.device("cuda:0"))
     >>> set_device("mps")
     >>> set_device("cpu")
+    >>> set_device("gpu")
     """
     set({"device": dev})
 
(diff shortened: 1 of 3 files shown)


gvarnavi:
Thanks @arthurmccray. 

On my phone now so can’t check, but perusing the files changed it seems to me this always falls back to cuda. The previous behavior was setting cuda/mps according to which modules are available.

arthurmccray:
Is that expected behavior? In `torch` the `"mps"` backend is treated as distinct, though admittedly `"gpu"` isn't an option either as you use `"cuda"` for the default GPU specification. 

So I guess the question is if we should use `"gpu"` as a default "accelerator" that quietly defaults to `"cuda"` if there are both mps and cuda available, or if we should ask people to specify `"mps"` if that's what they want.

gvarnavi:
Not sure what you mean haha? 😅 The mps and cuda backends are mutually exclusive. 

I'm saying users likely want to simply specify "cpu" or "gpu" and get acceleration on their respective hardware.

arthurmccray:
ah--i thought mps was for the integrated gpu, and didn't realize you couldn't have a device with an additional nvidia gpu as well (don't have any apple products lol). In any event I added `"mps"` as a fallback for when the user selects `"gpu"` and cuda is not available.

gvarnavi (approved):
Small testing bug on macos-latest. I suspect this would've been caught on PR to main, but we could discuss expanding the dev PR checks too (see https://github.com/electronmicroscopy/quantem/pull/142)

PR #167, opened 2026-01-31
Tomography Module

cedriclim1:
# What does this PR do?

~~**Tutorial notebooks to be written in https://github.com/electronmicroscopy/quantem-tutorials/tree/tomography_tutorials, reorganizing `conventional_tomography.py` and implementation of `AutoSerialize` needed, do not review yet - 1/30**~~
**Tutorials ready; still have some cleanup to do 2/02**
Implementation of conventional (simutlaneous iterative reconstruction technique (SIRT) and filtered back projection (FBP)) and machine-learning enabled (implicit neural representations) tomographic reconstruction methods. 

The code implements the overall design patterns of `diffractive_imaging` i.e, usage of `core/ml` classes, and reconstruction loop design. Hopefully everything is still able to be easily extended to different types of tomography experiments. 

Since this a big PR, I'll provide a brief description of the relevant `.py` files in `tomography`.

-  `tomography.py`: Top-level file that contains the reconstruction loop and instantiation of the `Tomography` object through `.from_models.`
- `tomography_lite.py`: Similar to `ptychography_lite.py`, this file abstracts the object, model, optimizer, and scheduler initialization to simply loading the tomographic dataset and perform an immediate reconstruction.
- `tomography_base.py`: Base class that inherits from `AutoSerialize`, `RNGMixin`, and `DDPMixin` (*new*) with the appropriate properties that is needed for every reconstruction.
- `tomography_opt.py`: Contains all the necessary optimizer parameters for reconstructions i.e, object and pose. 
- `object_models.py`: Contains the classes for both pixelated and INR reconstructions. Can directly pretrain the volume from conventional methods.
- `dataset_models.py`: Contains pixelated and INR datasets with their respective `.forward` and `.__getitem__` calls. 
- `logger_tomography.py`: Contains the logger for tomography reconstructions.
- `utils.py`: Various functions for helping process tilt series datasets and also has the tools for performing voxelwise AD reconstructions.

There are also some added functionalities to `core/ml` that was implemented for helping initialization of distributed computing on HPC platforms (NERSC). The updates included briefly described here:
- `core/ml/ddp.py`: Initializes all the necessary parameters needed for doing distirbuted computing (defining world size, global rank, and local rank). Also contains helper functions for setting up model parallelization using `DistributedDataParallel`and setting up `DataLoader` distributed sampling using `DistributedSampler`
- `core/ml/inr.py`: Added [Winner initialization ](https://arxiv.org/pdf/2509.12980) to SIREN neural networks.
- `core/ml/loss_functions.py`: Added custom loss functions as `nn.Module`'s.
-  `core/ml/profiling.py`: Context manager for profiling code using NVIDIA Nsight.
# What should the PR reviewer do?
The main points to check for this PR would be:
- [ ] Ensure that all example scripts and notebooks are easily run using local or HPC clusters.
- [ ] Check all changes in `core/ml` priority to `loss_functions.py` if we want to turn loss functions into `nn.Module`'s. 
- [ ] `object_models.py` and `dataset_models.py` follow the same design patterns and does not contain any redundant code that might be inherited from base classes.
- [ ] Check the reconstruction loop in `tomography.py` and check if loss calculations are being performed correctly, i.e stepping the scheduler/optimizer at the correct places.

Please make note of any design patterns that were not followed or potential bugs. 

@arthurmccray will notify you when examples are ready to be tested.


cedriclim1:
@arthurmccray tutorials are ready: https://github.com/electronmicroscopy/quantem-tutorials/tree/tomography_tutorials

src/quantem/core/ml/ddp.py
```diff
@@ -0,0 +1,183 @@
...
+            print(f"  Global batch size: {batch_size * self.world_size}")
+            print(f"  Train batches per GPU per epoch: {len(train_dataloader)}")
+
+            if val_dataset:
+                print(f"  Total val samples: {len(val_dataset)}")
+                print(f"  Val batches per GPU per epoch: {len(val_dataloader)}")
+
+        return train_dataloader, train_sampler, val_dataloader, val_sampler
+
+    def build_model(
```
arthurmccray: shouldn't this be `distribute_model` or similar? Building the model implies initialization I think.

src/quantem/core/ml/ddp.py
```diff
@@ -0,0 +1,183 @@
...
+        """
+        Wraps the model with DistributedDataParallel if mulitple GPUs are available.
+
+        Returns the model.
+        """
+        print("Building Model on device: ", self.device)
+        model = model.to(self.device)
+        if pretrained_weights is not None:
+            model.load_state_dict(pretrained_weights.copy())
+
```
arthurmccray: Why is loading the weights done here? Feels rather independent of `ddp`

src/quantem/core/ml/constraints.py
```diff
@@ -0,0 +1,99 @@
...
+from typing import Any, Self
+
+import numpy as np
+import torch
+
+
+@dataclass(slots=True)
+class Constraints(ABC):
+    """
+    Needs to be implemented in all object models that inherit from BaseConstraints.
```
arthurmccray: I would maybe add a slightly extended docstring here. The idea is that any Model that inherits from `BaseConstraints` also has  `.constraints` attribute that is of type `Constraints`, right?

src/quantem/core/ml/constraints.py
```diff
@@ -0,0 +1,99 @@
...
+    DEFAULT_CONSTRAINTS = Constraints()
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._soft_constraint_losses = []
+        self.constraints = self.DEFAULT_CONSTRAINTS.copy()
+
+    @property
+    def soft_constraint_losses(self) -> list[float]:
+        return np.array(self._soft_constraint_losses)
```
arthurmccray: type hinting of this is inconsistent

src/quantem/core/ml/inr.py
```diff
@@ -109,6 +110,15 @@ def _build(self) -> None:
         net_list.append(self._final_activation)
         self.net = nn.Sequential(*net_list)
 
+        if self.winner_initialization:
+            with torch.no_grad():
+                self.net[0].linear.weight += (
+                    torch.randn_like(self.net[0].linear.weight) * 5 / self.first_omega_0
+                )
+                self.net[1].linear.weight += (
+                    torch.randn_like(self.net[1].linear.weight) * 0.1 / self.hidden_omega_0
```
arthurmccray: do you think we should have a way of seeding the initialization? Training isn't fully deterministic, but it's generally a good idea to do what we can. One (inelegant) way would be to allow `winner_initialization: bool | int` and if an `int` is passed, use that to seed a `torch.Generator`.

src/quantem/core/ml/loss_functions.py
```diff
@@ -139,3 +141,98 @@ def combined_l2(pred: torch.Tensor, target: torch.Tensor, alpha: float = 0.7) ->
...
+    def __init__(
+        self,
+        reduction: str = "mean",
+    ):
+        super(MSELoss, self).__init__()
+        self.reduction = reduction
+
+    def forward(self, pred: torch.Tensor, target: torch.Tensor) -> torch.Tensor:
+        return torch.nn.functional.mse_loss(pred, target, reduction=self.reduction)
+
```
arthurmccray, line 108: Doesn't this almost exactly re-implement `torch.nn.MSELoss` and `torch.nn.L1Loss`? I'm not sure I see the point of having it.

src/quantem/core/ml/loss_functions.py
```diff
@@ -139,3 +141,98 @@ def combined_l2(pred: torch.Tensor, target: torch.Tensor, alpha: float = 0.7) ->
     comp_l2 = complex_l2(pred, target)
     amp_ph_l2 = amp_phase_l2(pred, target)
     return alpha * amp_ph_l2 + (1 - alpha) * comp_l2
+
+
+# TODO: Better loss function implementation? More torch-like.
```
arthurmccray: Well, `torch` has all it's losses as both modules and functions and I suggest we follow along with that convention to a point, though it can get confusing. Mainly though I think as long as we're consistent internally then it's fine.

So if we're wanting to have all our losses be defined as `nn.Modules` as done below, then we should also update `get_loss_function` to return a module as well. Does that make sense? 

The one thing I'm not sure about is if we should write our own functional versions of the losses (though I think probably not?), i.e. if there should be a `llmse_loss()` function somewhere that is called by `LLMSELoss`, to be consistent with the pytorch implementation of `torch.nn.MSELoss` calling `torch.nn.functional.mse_loss`

src/quantem/core/ml/profiling.py
arthurmccray, line 1: thanks for adding this!

src/quantem/core/utils/imaging_utils.py
```diff
@@ -1,14 +1,22 @@
 # Utilities for processing images
 
 import math
-from typing import Optional, Tuple
+from typing import Any, Optional, Tuple
 
 import numpy as np
 import torch
+
+# TODO: Temporary
```
arthurmccray: intended to be left here?

src/quantem/core/utils/imaging_utils.py
```diff
@@ -1050,118 +1058,157 @@ def unwrap_phase_2d_torch(
         )
 
 
-def radially_project_fourier_tensor(
```
arthurmccray, line 1053: was this function totally removed or just moved to a different file? and presumably someone has checked to make sure it doesn't break things in the `imaging` module?

src/quantem/core/utils/imaging_utils.py
arthurmccray, line 844: let's use the standard `# type:ignore` syntax as not everyone uses `ty`.

src/quantem/core/utils/imaging_utils.py
```diff
@@ -1050,118 +1058,157 @@ def unwrap_phase_2d_torch(
...
+
+def background_subtract(
+    image: ImageType,
+    mask: Optional[BoolArray] = None,
+    thresh_bg: Optional[float] = None,
+    order: Tuple[int, int] = (1, 1),
+    sigma: Optional[float] = None,
+    num_iter: int = 10,
+    plot_result: bool = True,
+    axsize: Tuple[int, int] = (3.1, 3),
```
arthurmccray: default `float` for specified `int`

src/quantem/core/utils/imaging_utils.py
arthurmccray, line 1: are the changes made here reflected in the imaging branch? Dangerous to be changing the utils for a different module, as that could cause things to break there, or changes made in imaging pulled into dev could break tomo.

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,547 @@
...
+        learn_shift: bool = True,
+        learn_tilt_axis: bool = True,
+        _token: object | None = None,
+    ):
+        AutoSerialize.__init__(self)
+        OptimizerMixin.__init__(self)
+        nn.Module.__init__(self)
+        # if _token is not self._token: TODO: Idk why this isn't working.
+        #     raise RuntimeError("Use TomographyPixDataset.from_* to instantiate this class.")
+
```
arthurmccray, line 179: should probably fix this--refer to `Dataset` implementation for an example if needed

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,547 @@
...
+        self._learnable_tilts = learnable_tilts
+
+    @property
+    def params(self) -> dict[str, torch.nn.Parameter]:
+        """
+        Returns the parameters that should be optimized for this dataset.
+
+        Should be implemented in subclasses.
+        """
+        raise NotImplementedError("This method should be implemented in subclasses.")
```
arthurmccray: Does this actually get used? I don't see it being overloaded in `TomographyPixDataset` and in `TomographyINRDataset` it just returns the `nn.Module` builtin `self.parameters()`

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,547 @@
...
+        return list[nn.Parameter](self.params)
+
+
+class TomographyPixDataset(TomographyDatasetBase):
+    """
+    Dataset class for pixel-based tomography, i.e AD, SIRT, WBP, etc...
+
+    TODO:
+     - What should the forward pass return? In AD, it should be both the tilt image and the pose.
+       In SIRT, it's only the tilt image.
```
arthurmccray: should probably update docstrings

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,547 @@
...
+
+        pixel_i = remaining // self.tilt_stack.shape[1]
+        pixel_j = remaining % self.tilt_stack.shape[1]
+
+        # return DatasetValue(
+        #     target=self.tilt_stack[projection_idx, pixel_i, pixel_j],
+        #     tilt_angle=self.tilt_angles[projection_idx],
+        #     projection_idx=projection_idx,
+        #     pixel_loc=(pixel_i, pixel_j),
+        # )
```
arthurmccray: necessary?

src/quantem/tomography/dataset_models.py
arthurmccray, line 1: are there constraints for the `Dataset` learned params such as `z1`?

src/quantem/tomography/object_models.py
```diff
@@ -1,540 +1,630 @@
...
+        device: str = "cpu",
+        rng: np.random.Generator | int | None = None,
     ):
         super().__init__(
-            volume_shape=volume_shape,
+            shape=shape,
             device=device,
+            rng=rng,
+            _token=self._token,
         )
```
arthurmccray, line 298: we're trying to encourage using classmethods for all instantiation. This could be `ObjectPixelated.from_uniform()`, which would be more descriptive and allows for `ObjectPixelated.from_array()` as well (e.g. to instantiate with the ground truth to ensure that the solution isn't diverging).

src/quantem/tomography/object_models.py
```diff
@@ -1,540 +1,630 @@
...
-            self._pretrain_lrs = []
-
-        if model_input is not None:
-            self.model_input = model_input
-
-        if pretrain_target.shape[-3:] != self.model_input.shape[-3:]:
-            raise ValueError(
-                f"Pretrain target shape {pretrain_target.shape} does not match model input shape {self.model_input.shape}"
+            raise NotImplementedError(
+                "TODO: Resseting the model to the pretrained weights is not implemented yet. To make this work I would have to reinstantiate the model I think."
```
arthurmccray: I don't think so--you should be able to just reset the model weights (most of the models I have built, including `inr` have a `.reset()` method. Or, alternatively and perhaps better, you could just grab the original model weights when you instantiate the `ObjectINR.from_model()`, and then set the state dict to them here.

src/quantem/tomography/tomography.py
```diff
@@ -1,237 +1,481 @@
-import torch
+from typing import List, Literal, Optional, Self, Tuple
```
arthurmccray: Prefer 3.11+ type hinting, so just use `list` and `tuple` rather than `List` and `Tuple`

src/quantem/tomography/tomography.py
```diff
@@ -1,237 +1,481 @@
...
+        reset_dset: DatasetModelType | None = None,
     ):
-        super().__init__(dataset, volume_obj, device, _token)
+        """
+        This function should be able to handle both AD and INR-based tomography reconstruction methods.
+        I.e, auto-detection through the obj model type, while both share the same pose optimization.
+        """
 
-    # --- Reconstruction Method ---
+        # TODO: Prior to reconstruction, it is assumed that object + dataset are both in the correct devices. Need to implement a way to check this.
```
arthurmccray: or could just send them to the device again, I don't think that would cause any problem

src/quantem/tomography/tomography.py
arthurmccray, line 1: Do you intend/want to leave all of the profiling in for pulling to dev? I would suggest not--have a separate `tomography-profiling` branch that you can just pull into from `dev` or `tomography` when changes are made. It's a little annoying because I'm not sure if cherry picking the commits/improvements from the `tomography-profiling` branch back into `tomography` would work... but I don't love leaving all the profiling mess in `dev`.

src/quantem/tomography/tomography.py
```diff
@@ -1,237 +1,481 @@
...
+                                num_samples_per_ray=curr_num_samples_per_ray,
+                                val_loss=avg_val_loss if self.val_dataloader is not None else None,
+                            )
+
+                        self.logger.flush()
+
+    # --- Helper Functions ---
+
+    def save_volume(self, path: str = "recon_volume.npz"):
+        # TODO: Temporary, need to talk to Arthur what the correct way of saving results is.
```
arthurmccray: Honestly I would suggest we have the method for saving the full `Tomography` with `AutoSerialize`, but that we don't have a method for saving the volume. Instead encourage users to do so however they want. So in the tutorials just show you can access `tomo.obj` and the aligned shifts and such, and demo saving them with a simple `np.save()`

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,547 @@
...
+            learn_shift=learn_shift,
+            learn_tilt_axis=learn_tilt_axis,
+        )
+
+    # --- Optimization Parameters ---
+    # @property
+    # def params(self):
+    #     # TODO: Need to double check if this is correct way, also need to implement get_optimization_parameters @Arthur!!
+    #     raise NotImplementedError("This method should be implemented in subclasses.")
+
```
arthurmccray, line 240: I don't remember exactly what we decided when we chatted about this--but as long as it's consistent with what you're doing in your object models i'm fine with it

src/quantem/tomography/tomography_lite.py
```diff
@@ -0,0 +1,151 @@
...
+
+class TomographyLiteINR(Tomography):
+    """
+    A lite version of the Tomography class.
+    """
+
+    @classmethod
+    def from_dataset(
+        cls,
+        dset: DatasetModelType,
```
arthurmccray: similar to `TomographyPixLite` I would suggest instantiating the `TomgraphyINRDataset` here rather than having the user do so. It seems like that is just an empty initialization passing the tilt series and angles.

src/quantem/tomography/utils.py
arthurmccray, line 1: what do you think about moving some of the functions from `/tomography/utils.py` to `core/utils/tomography_utils.py`? That would expose some of these functions to other modules (such as cross correlation), and other functions (such as the filtering) could be moved to `core/utils/filter.py`

arthurmccray (commented):
overall looks good! The tutorial notebooks all run, though they do need some additional explanations and I left some comments at the PR https://github.com/electronmicroscopy/quantem-tutorials/pull/4 

Let me know if you have any issues or questions with my comments!

src/quantem/core/ml/constraints.py
cedriclim1: See commit 46dda7c, docstring has been updated. Not quite sure what else to add.

src/quantem/core/ml/constraints.py
cedriclim1: See commit 46dda7c, type hinting has been updated.

src/quantem/core/ml/ddp.py
cedriclim1: See 5c617d3, renamed to distribute model.

src/quantem/core/ml/ddp.py
cedriclim1: See 5c617d3, `def reset` within `object_models.py` now loads the state dict from the pretrained weights and redistributes the model.

src/quantem/core/ml/inr.py
cedriclim1: See [9465aa7](https://github.com/electronmicroscopy/quantem/pull/167/commits/9465aa772da4f9099a3fd3414dc37b600b03dada), included a `rng=torch.Generator()` let me know if this works or not...Didn't see a direct comparison in your code? If there is please feel free to reference it here.

src/quantem/core/ml/loss_functions.py
cedriclim1: I don't think I have a strong opinion either way. I just find the `nn.Module` way of handling loss functions just to be more cleaner especially if there are multiple parameters in the loss function. For example, `CharbonnierLoss` has an `epsilon` and `reduction` term that can be easily retrieved doing `CharbonnierLoss.epsilon`, etc... Especially as we move towards intricate loss functions, this might prove useful.

src/quantem/core/ml/loss_functions.py
cedriclim1: Agreed see commit [953ee94](https://github.com/electronmicroscopy/quantem/pull/167/commits/953ee9443755f2a9c616f7b240655307f6d27316), I don't know why I have this here either lol.

src/quantem/core/utils/imaging_utils.py
cedriclim1: Ok, so I don't know why this was totally removed. This Colin's original implementation of background subtraction, that was just either moved or removed. I can't find it any of the `utils.py` funcftions so I have just readded it back in. Happy to talk about this more.

cedriclim1: Just double checked https://github.com/electronmicroscopy/quantem/tree/imaging, `background_subtraction` is completely gone as well.

src/quantem/core/utils/imaging_utils.py
cedriclim1: Function was totally removed, again totally separate from the `imaging` branch.

src/quantem/core/utils/imaging_utils.py
cedriclim1: Put temporary in there since these imports are for `background_subtraction`. Can talk about this in person.

src/quantem/core/utils/imaging_utils.py
cedriclim1: See commit [bde606f](https://github.com/electronmicroscopy/quantem/pull/167/commits/bde606fd48f70dd3471df0d4b201a12c20b263c9), fixed here.

src/quantem/core/utils/imaging_utils.py
cedriclim1: See commit [bde606f](https://github.com/electronmicroscopy/quantem/pull/167/commits/bde606fd48f70dd3471df0d4b201a12c20b263c9), fixed here.

src/quantem/tomography/dataset_models.py
cedriclim1: I'm not sure what you mean by this. Are you talking about a total-variational loss for the `z1` tensor for example? If so, no. If you think it's worth reimplementing it I can do this. I've actually run into an issue where if the dataset is too poorly constrained, the higher tilt angle Euler angles start to blow up. Not sure if including a TV-loss to that would help.

src/quantem/tomography/dataset_models.py
cedriclim1: Fixed, see commit [c0b61a3](https://github.com/electronmicroscopy/quantem/pull/167/commits/c0b61a36511557d6ba90c37a11575ab7ed2c33f5)

src/quantem/tomography/dataset_models.py
cedriclim1: See commit [d42611b](https://github.com/electronmicroscopy/quantem/pull/167/commits/d42611bddfd7bc7a5c6f01e57a996249dd284c91) and [aea34b1](https://github.com/electronmicroscopy/quantem/pull/167/commits/aea34b1b19421b1b41b812fd1b211ba78de30a9d), and [038e94d](https://github.com/electronmicroscopy/quantem/pull/167/commits/038e94d4d1267ec7a5ca92096e3caf7c2784ff87) (Apologies just some type hinting stuff that I forgot to do in the previous 2 commits lol).

I remember we had a conversation where `OptimizerMixin` only works if `get_optimization_parameters` is defined. In this case here, the workaround I did was to have `get_optimization_parameters` just return a list of parameters from `params`.

src/quantem/tomography/dataset_models.py
cedriclim1: OK, I see why I did it this way. It all has to do with `get_optimization_parameters` requiring the `self.parameters()`. But tbf, I don't know why I still have this thing, I'm pretty sure I can just call `self.parameters()` in the `get_optimization_parameters`. I know for some reason, defining `params` is still important for the actual `object_models.py`.

See [edab112](https://github.com/electronmicroscopy/quantem/pull/167/commits/edab1120c639b831ba7c139123d277400d2605cc) for the implementation. Also tested that the auxiliary parameter gradients were still being tracked correctly.

src/quantem/tomography/dataset_models.py
cedriclim1: See [888e6b9](https://github.com/electronmicroscopy/quantem/pull/167/commits/888e6b946e441b6001beb7942790d287878e49da). Not sure what else to add.

src/quantem/tomography/dataset_models.py
cedriclim1: See [c0a8777](https://github.com/electronmicroscopy/quantem/pull/167/commits/c0a8777409282cffe8b3f9a2a86dd165b803314d). 

The original thought of having the `DatasetValue` was cool in theory, but putting dataclasses directly into the `__getitem__` makes torch very upsetty.

src/quantem/tomography/utils.py
cedriclim1: I actually like this idea. This could mean I can combine the `imaging_utils.py` stuff for background subtraction to be added into `tomography_utils.py` without having to interefere with other feature branches.

src/quantem/tomography/tomography.py
cedriclim1: All the profiling stuff have been removed, see commit [c4c6e8a](https://github.com/electronmicroscopy/quantem/pull/167/commits/c4c6e8a65f2c318ae07e1f68a564b20c08c647a2). 

I think I'll end up having a `tomnography-profiling` branch, but I think this can be put somewhere in my fork or something.

src/quantem/tomography/tomography.py
cedriclim1: Well `AutoSerialize` already does work with `Tomography` with saving the whole thing. I think this is just particularly useful for if you want to log the volume at a specific epoch or something. In general, I think most users would just prefer to have direct access to the finished volume in `.npy` or `.pt` format so that they can play around with in TomViz or whatever. 

I also do see your point where in the logging I can just directly call `tomo.obj` and save from there...But I think it might be just fool-proof to have a `save_volume` function.

cedriclim1:
Ok @arthurmccray, most comments have been resolved as of 2/23/2026. There's some things that I have to look at on your comments on `object_models.py` and some general comments on the top-level classes. The next focus here will be getting `TomographyLogger` back to working since I do have the additional channel dimension for creating the volume `[C, N, N, N]`. There's also a ton of linting problems to fix.

src/quantem/core/ml/inr.py
arthurmccray: yeah I think this is fine--the other option would be to inherit from `RNGMixin` but that feels like overkill in this case. I think what you have is better.

src/quantem/core/utils/imaging_utils.py
arthurmccray: yeah this wasn't you haha 😆

src/quantem/core/utils/imaging_utils.py
arthurmccray: yay linters!

src/quantem/core/utils/imaging_utils.py
arthurmccray: Okay looks good, @cophus or @darshan-mali  will this cause any issues with the imaging branch?

src/quantem/tomography/dataset_models.py
arthurmccray: yeah it's mostly we shouldn't leave TODOs around if they're completed

src/quantem/tomography/dataset_models.py
arthurmccray: Yeah it seems like, as implemented, there isn't a way of doing soft constraints for the datasets. I think we might want to include that here (e.g. TV for z1 like you mention) even if it's not going to be used much, because then we'll have a template for how to implement soft constraints in the more advanced tomography dsets (through-focal, real-probe, hyperspectral) for which soft constraints will probably be necessary.

src/quantem/tomography/tomography.py
arthurmccray: okay that seems reasonable. In that case please: 
- make sure the documentation is explicit that it's not saving the full `Tomography` class
- allow for an overwrite flag
- decide if you want to handle saving as `torch` or `numpy` or just `numpy`, but either way make it explicit in the docstring

src/quantem/tomography/object_models.py
cedriclim1: See commit [6d9cbf6](https://github.com/electronmicroscopy/quantem/pull/167/commits/6d9cbf6dd3a4891f6b7feb3a8b9b41e49b8d44cc), seems like I forgot to do it only for this Object.

src/quantem/tomography/object_models.py
cedriclim1: See commits [ddd473d](https://github.com/electronmicroscopy/quantem/pull/167/commits/ddd473d02aa5fb39b6a953cebafcf7126391b7c4) and [5c617d3](https://github.com/electronmicroscopy/quantem/pull/167/commits/5c617d3521bb55c70dc788884461c30f636b7033), `def reset()` was implemented along with renaming of `distribute_model`. Basically just loading the pretrained weights again into the model.

src/quantem/tomography/tomography.py
cedriclim1: See [d51aa51](https://github.com/electronmicroscopy/quantem/pull/167/commits/d51aa51336aea67134f3bd0a482f1ecd404dbd5d).

src/quantem/tomography/tomography.py
cedriclim1: See [d51aa51](https://github.com/electronmicroscopy/quantem/pull/167/commits/d51aa51336aea67134f3bd0a482f1ecd404dbd5d)

src/quantem/tomography/tomography_lite.py
cedriclim1: See [16aa0f2](https://github.com/electronmicroscopy/quantem/pull/167/commits/16aa0f222e9f1ce6fb58fb9f6eff675a68534d68). 

`.from_dataset` only now requires the tilt_series and tilt_angles.

cedriclim1 (commented):
misclick

src/quantem/tomography/utils.py
cedriclim1: See [945064e](https://github.com/electronmicroscopy/quantem/pull/167/commits/945064e2b5b6b5c0df284d0f3db15b3452bff52c). 

Moved most utility functions to `tomography_utils.py`. `background_subtraction` is also moved off to `tomography_utils.py`. Also put Gaussian kernel stuff in `filters.py`.

src/quantem/tomography/tomography.py
cedriclim1: See commit [c2b7d26](https://github.com/electronmicroscopy/quantem/pull/167/commits/c2b7d264297ff7f7596e350c1b315eb04acf2862). 

Overwrite flag included, and a small docstring on what this function actually does.

src/quantem/core/utils/imaging_utils.py
cedriclim1: Resolved, included in `tomography_utils.py`

src/quantem/core/utils/imaging_utils.py
arthurmccray: -- no longer relevant cuz of `tomography_utils.py`

src/quantem/core/ml/loss_functions.py
cedriclim1: See commit [446a6da](https://github.com/electronmicroscopy/quantem/pull/167/commits/446a6da05edbbed5b00de6339236e0073bbc4c3f).

Everything has been changed to modules, I've also renamed the symbol for `get_loss_function` --> `get_loss_module`. This should have covered all the Ptycho stuff, but please double check.

arthurmccray: Cool looks good. You could include your new loss terms in the `get_loss_module` if you'd like (and just pass kwargs), but it's not particularly important as I really want that function to handle the more normal losses.

arthurmccray:
Overall looks good! I see the remaining things being: 

- [x] Finish the required channel stuff (maybe already done?) 
- [x] optional but preferred, add the infrastructure for constraints in the datasets. We could also punt this to a point when we really need it--e.g. as part of the real-probe stuff--but it might be nice to have for processing this next batch of experiments.
- [x] updating the tutorials to reflect these changes, and also respond to comments on the tutorials PR: https://github.com/electronmicroscopy/quantem-tutorials/pull/4

Once these are done I will go through all the notebooks again as well as testing the hpc scripts on nersc, and then it should be done!

cedriclim1:
@arthurmccray Please look at commit [bfc6e06](https://github.com/electronmicroscopy/quantem/pull/167/commits/bfc6e062365d5613a4b8dc38c6a6c6d222553a72). There are some functions that I had to do `pyright ignores`. For example: `get_tv_loss` is an abstract method in the class `ObjectBase`. However, `ObjectPixelated` and `ObjectINR` both have different implementations of `get_tv_loss`, so the base method would have `**kwargs` which gave linting errors with the overloaded methods with different keyword args.

src/quantem/tomography/dataset_models.py
cedriclim1: Made a first iteration of this by adding some soft constraints in the dataset models, see [9ecd8d7](https://github.com/electronmicroscopy/quantem/pull/167/commits/9ecd8d7aeaec14d81a5d5cb805ef02723e53c449). I think there has to be a better way of handling all these constraints together rather than disjointly in `object_models.py` and `dataset_models.py`. Will probably need some time to think about it, but for now this implementation works for soft constraints on the dataset.

cedriclim1:
@arthurmccray I think I addressed your 3 comments. Please look at the tutorials again, as they have been updated in addition to some minor changes in the files here due to issues in the tutorials.

cedriclim1:
@arthurmccray it seems all the issues we were running into last week stemmed from the DataLoader instantiation and not passing a `multiprocessing_context` variable. In the Python docs: https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods. In PyTorch DataLoader docs: https://docs.pytorch.org/docs/stable/data.html

There are three start methods: `spawn`, `fork`, and `forkserver`. I still need to take a deeper look at this, but for now in `core/ml/ddp.py`'s `setup_dataloader` function, I have set the `multiprocessing_context="spawn"`. This might not be the best for POSIX systems which should default to `"forkserver"` but this seems to be causing some errors.

*Edit: Seems like `spawn` is safer especially if we want this to be runnable across different OS's. I'll maybe open up an issue on doing PyTests for DDP related stuff.*

src/quantem/core/ml/ddp.py
arthurmccray, line 1: Are there any end-of-task items that would justify a `cleanup` method for `DDPMixin`? e.g. cleaning up any `torch.mp` spawned processes.

src/quantem/tomography/tomography_base.py
```diff
@@ -1,499 +1,137 @@
...
+        self.logger = logger
+
+        # Loss
+        self._epoch_losses: list[float] = []
+        self._consistency_losses: list[float] = []
+        self._val_losses: list[float] = []
+        # DDP Initialization
+        if isinstance(obj_model, ObjectINR):
+            print("Setting up DDP for obj_model")
+            self.setup_distributed(device=device)
```
arthurmccray, line 55: This print should probably be only on global_rank 0, right? Currently prints once per gpu.

src/quantem/tomography/tomography.py
```diff
@@ -1,237 +1,548 @@
...
+        N = max(self.obj_model.shape)
 
-            pbar.set_description(f"SIRT Reconstruction | Loss: {loss.item():.4f}")
+        if num_samples_per_ray is None:
+            num_samples_per_ray = max(self.obj_model.shape)
+        else:
+            if isinstance(num_samples_per_ray, int):
+                num_samples_per_ray = num_samples_per_ray
+            else:
+                print("num_samples_per_ray schedule provided.")
```
arthurmccray: This also prints once per gpu

src/quantem/core/ml/inr.py
```diff
@@ -109,6 +110,21 @@ def _build(self) -> None:
...
 
+        if self.winner_initialization:
+            if type(self.winner_initialization) is int:
+                rng = torch.Generator()
+                rng.manual_seed(self.winner_initialization)
+            else:
+                rng = torch.Generator()
+                rng.manual_seed(42)
+            with torch.no_grad():
+                self.net[0].linear.weight += (  # pyright: ignore[reportAttributeAccessIssue]
```
arthurmccray: Since we're not specifying that everyone use the same linters, let's use the more generic `# type: ignore ...` (since I don't know if `ty` or others would respect `# pyright: ignore ...`)

src/quantem/core/utils/filter.py
```diff
@@ -109,3 +110,53 @@ def otsu_threshold(img: np.ndarray, bins: int = 256) -> float:
...
+def gaussian_filter_2d(
+    img: torch.Tensor, sigma: float, kernel_1d: torch.Tensor
+) -> torch.Tensor:  # Add kernel_1d as an argument
+    # kernel_1d = gaussian_kernel_1d(sigma)  # Create 1D Gaussian kernel - Moved outside function
+    padding = len(kernel_1d) // 2  # Ensure that image size does not change
+    img = img.unsqueeze(0).unsqueeze_(0)  # Make copy, make 4D for ``conv2d()``
+    # Convolve along columns and rows
+    img = torch.nn.functional.conv2d(img, weight=kernel_1d.view(1, 1, -1, 1), padding=(padding, 0))
+    img = torch.nn.functional.conv2d(img, weight=kernel_1d.view(1, 1, 1, -1), padding=(0, padding))
+    return img.squeeze_(0).squeeze_(0)  # Make 2D again
```
arthurmccray, line 135: This is fine for now I think. Eventually we might want to have this be a little more robust so that it can be used by other modules, mainly in the sense of being n-dimensional like [scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.ndimage.gaussian_filter.html) but for torch. But that can be done whenever we have a use case for it.

src/quantem/core/utils/tomography_utils.py
```diff
@@ -0,0 +1,316 @@
...
+
+def _as_array(x: ImageType) -> NDArray[Any]:
+    return (
+        x.array
+        if isinstance(
+            x,
+        )
+        else np.asarray(x)
+    )
+
```
arthurmccray: We have the `to_numpy()` function `/core/utils/utils` that does this, also as written this should be caught by a linter cuz it would fail I think?

src/quantem/core/utils/tomography_utils.py
```diff
@@ -0,0 +1,316 @@
...
+def differentiable_shift_2d(image, shift_x, shift_y, sampling_rate):
+    """
+    Shifts a 2D image using grid_sample in a differentiable manner.
+
+    Args:
+        image: Tensor of shape [H, W]
+        shift_x: Scalar tensor (dx) for shift in x-direction (in physical units)
+        shift_y: Scalar tensor (dy) for shift in y-direction (in physical units)
+        sampling_rate: Scalar value (physical units per pixel) to correctly normalize shifts
+
```
arthurmccray, line 240: should say in docstring the boundary conditions applied

src/quantem/tomography/tomography.py
arthurmccray, line 1: Somewhere please add a `.plot_losses()` or `.visualize()` method. Could be in `tomography` or if you want to have a `tomography_viz.py` that would work too.

src/quantem/tomography/logger_tomography.py
```diff
@@ -0,0 +1,90 @@
...
+class LoggerTomography(LoggerBase):
+    """
+    Logger for ML-based tomography reconstructions.
+    """
+
+    def __init__(
+        self,
+        log_dir: str,
+        run_prefix: str,
+        run_suffix: str = None,
```
arthurmccray: `str | None`

src/quantem/tomography/dataset_models.py
```diff
@@ -0,0 +1,687 @@
...
+        if not (
+            tilt_stack.shape[0] < tilt_stack.shape[1] or tilt_stack.shape[0] < tilt_stack.shape[2]
+        ):
+            raise ValueError(
+                "The number of tilt projections should be in the first dimension of the dataset."
+            )
+
+        # TODO: Maybe have the validation in here too.
+        max_val = np.quantile(tilt_stack, 0.95)
```
arthurmccray: your linter should catch this one; `tilt_stack` could be a `Dataset3d` or `Tensor` and the `np.quantile(...)` would error.

src/quantem/tomography/tomography.py
```diff
@@ -1,237 +1,548 @@
...
-        device,
-        _token,
+        num_iter: int = 10,
+        batch_size: int = 1024,
+        num_workers: int = 32,
+        reset: bool = False,
+        optimizer_params: dict | None = None,
+        scheduler_params: dict | None = None,
+        obj_constraints: dict | ObjConstraintsType | None = None,
+        dset_constraints: dict | DatasetConstraintsType = None,
```
arthurmccray: ` | None`

src/quantem/tomography/tomography.py
arthurmccray, line 1: I'm seeing quite a few linter errors here (and on some of the other files), but i'm only pointing out the easy to fix ones. I'll leave it up to you how much you want to clean them up, but we could sit down and fix them together if you'd like.

arthurmccray (commented):
overall looks good! Left a few comments, but it's almost there :) 
once these are fixed and https://github.com/electronmicroscopy/quantem-tutorials/pull/4 it should be ready to merge!

src/quantem/core/ml/ddp.py
cedriclim1: Yea we shouldn't do this as discussed earlier today. I.e, since DDP `DataLoader`'s are setup to have persistent workers, deleting the processes would lead to DDP being unhappy about it. This would probably be OK to do on HPC systems, but I think it's better to explicitly at the end of the script just call the `dist.destroy_all_processes()`

src/quantem/tomography/tomography_base.py
cedriclim1: Fixed! See [c2f46fe](https://github.com/electronmicroscopy/quantem/pull/167/commits/c2f46fed7999ec15babcc54c38a9155e137d7e9d)

src/quantem/tomography/tomography.py
cedriclim1: Fixed see here [b922cbb](https://github.com/electronmicroscopy/quantem/pull/167/commits/b922cbb458dfa53f5c8f5ce823efcc4943cccf22)

src/quantem/core/ml/inr.py
cedriclim1: Fixed here: [c81b2b5](https://github.com/electronmicroscopy/quantem/pull/167/commits/c81b2b5f5fff307477a58555ea947853599d4b5d)

src/quantem/core/utils/filter.py
cedriclim1: Sounds good, I'll put this on my to-do list.

src/quantem/core/utils/tomography_utils.py
cedriclim1: Fixed [62616d7](https://github.com/electronmicroscopy/quantem/pull/167/commits/62616d706700730c25768ef280a7dc7e7c599a57)

src/quantem/core/utils/tomography_utils.py
cedriclim1: Fixed [3567b4e](https://github.com/electronmicroscopy/quantem/pull/167/commits/3567b4e30a3d65ebdfea275df98857ac414e3a02)

src/quantem/tomography/tomography_base.py
cedriclim1: And here [0ec02df](https://github.com/electronmicroscopy/quantem/pull/167/commits/0ec02dfc4f021032485525fda48a826834212c1b)

src/quantem/tomography/logger_tomography.py
cedriclim1: Fixed [8afb028](https://github.com/electronmicroscopy/quantem/pull/167/commits/8afb02854d98f4bbd5592309395279f456674a7d)

src/quantem/tomography/tomography.py
cedriclim1: See here: [5923624](https://github.com/electronmicroscopy/quantem/pull/167/commits/5923624a7a275ba39411eacbb0bf888b4d1da8fd) and [b0171fd](https://github.com/electronmicroscopy/quantem-tutorials/pull/4/commits/b0171fdf7881d693a154f432c9f4cd9ce936bc7c)

src/quantem/tomography/dataset_models.py
cedriclim1: Fixed here [4ebb1d9](https://github.com/electronmicroscopy/quantem/pull/167/commits/4ebb1d9df4ee892afe862634838fa622e3632395)

src/quantem/tomography/tomography.py
cedriclim1: Fixed here [435e623](https://github.com/electronmicroscopy/quantem/pull/167/commits/435e623ed5198640449587b71a4733104bd4470a)

src/quantem/tomography/tomography.py
cedriclim1: Alright all the linting errors should be fixed (thanks Claude)! Also double-checked that the tutorials still run post fixing linter errors. Feel free to take a look at the commits with linting issues fixed.

cedriclim1:
@arthurmccray FYI - once this does get pushed into `dev`, did you make sure that `diffractive_imaging` will be fine with `get_loss_module`? I did find all instances of `get_loss_function` in both your `object_models.py` and `probe_models.py` but it might be worth checking if the tutorials still run. I can take a look tomorrow (3/24).

arthurmccray (approved):
yay

PR #169, opened 2026-02-02
MAPED data merging

cophus:
## What does this PR do?

This pull request adds support for multi angle precession electron diffraction (MAPED) data processing, described in this paper:  https://arxiv.org/abs/2506.11327

Typical usage is something like:

```
ds = []
for file in files:
    ds.append(
        em.core.io.read_4dstem(
            path + file,
            file_type = 'arina',
        )    
    )
maped = em.diffraction.MAPED.from_datasets(ds)
maped.preprocess();
maped.diffraction_origin()
maped.diffraction_align()
maped.real_space_align()
dataset_merged = maped.merge_datasets()
dataset_merged.save(
    file_output,
    mode = 'o',
)
```

Example workflow:

image

image

image

image

image

Notebook example applied to this data from @smribet https://drive.google.com/drive/folders/1EtNWeWZSO8TZ7Qibak2SVxQiAC9IsCCb?usp=sharing
[maped01.ipynb](https://github.com/user-attachments/files/24998301/maped01.ipynb)

## Instructions for reviewers

Please check this code for accuracy and test on other MAPED datasets. Likely the diffraction space origin finding will need some work - it uses a "real space biased cross correlation" which its the best solution I've come up with so far (pytest added for that function).

## TODO

- [x] Convert to torch
- [ ] (optional) speed up torch version?
- [ ] Test on other MAPED datasets
- [ ] Improve automated alignment for both real space and diffraction space.

### DIFF
--- a/src/quantem/diffraction/maped.py
+++ b/src/quantem/diffraction/maped.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from typing import Any, Sequence
 
 import numpy as np
+from scipy.signal.windows import tukey
 
 from quantem.core.datastructures.dataset4dstem import Dataset4dstem
 from quantem.core.io.serialize import AutoSerialize
@@ -34,7 +35,6 @@ class MAPED(AutoSerialize):
 
     def preprocess(
         self,
-        *,
         plot_summary: bool = True,
         scale: float | Sequence[float] | None = None,
         **plot_kwargs: Any,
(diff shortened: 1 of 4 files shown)


src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+                [[self.im_bf_merged, self.dp_mean_merged]],
+                title=[["Merged Bright Field", "Merged Mean Diffraction Pattern"]],
+                **plot_kwargs,
+            )
+
+        return dataset_merged
+
+
+def shift_images(
+    images,
```
cedriclim1, line 765: Should probably be moved to `core/imaging_utils.py`

ehrhardtkm:
When looking at experimental data with dead pixels (and not preprocessed data like above), be sure to filter them before `maped.diffraction_origin()`. 

Can fix with something like this:
```mask = ds[0].dp_mean.array > 1e4
em.visualization.show_2d(
    mask,
    vmax = 1.0,
)
mask = ds[0].dp_mean.array > 1e4
em.visualization.show_2d(
    mask,
    vmax = 1.0,
)

for d in ds:
    d.median_filter_masked_pixels(mask)
```

Attached is a notebook for a working example. Just replace the file names and path to run. [maped_test.ipynb](https://github.com/user-attachments/files/25026886/maped_test.ipynb)

src/quantem/diffraction/maped.py
cophus: Good point - I put it here for quick iteration, haven't moved it yet.

henrygbell:
I tried this code with my Si membrane MAPED data. 

Notebook attached here: [notebook] (https://github.com/user-attachments/files/25058505/HB_pr169_maped_test_nb.ipynb)
Data uploaded here if you are curious: [data] (https://drive.google.com/drive/folders/1CPl6Z0rKKleaQnG2Nkhj3ix8bZmh35nL?usp=drive_link)

This dataset has a lot of misalignment between the diffraction patterns in a single tilt, which this code does not deal with. I think we should add this functionality in the MAPED code base or the 4DSTEM dataset class (if it is not already).

image

Even with these misaligned DPs, the real space alignment worked well, I just needed around 15 alignment iterations.

Figure after maped.real_space_align()
image

cophus:
Thank you both for the testing! @henrygbell I should have clarified - MAPED stores the global diffraction shift, it doesn't apply shifts to the list of `MAPED.datasets`. So you need to look at the shifts afterwards 


image

```
from quantem.core.visualization import show_2d
rc = np.array(
    [
        [80, 80],
        [100, 80],
        [80, 100],
        [100, 100],
    ],
    dtype=int,
)

shifts_rc = -np.rint(maped.real_space_shifts).astype(int)
tiles = []
titles = []

for r, c in rc:
    row = []
    for i in range(5):
        dr, dc = shifts_rc[i]
        row.append(ds[i].array[r + dr, c + dc])
    row.append(dataset_merged.array[r, c])
    tiles.append(row)
    titles.append([f"ds[{i}] @ ({r+shifts_rc[i,0]},{c+shifts_rc[i,1]})" for i in range(5)] + [f"merged @ ({r},{c})"])

show_2d(
    tiles,
    title=titles,
    norm={
        "lower_quantile": 0.3,
        "upper_quantile": 0.999,
        "power": 0.5,
    },
    cmap="turbo_black",
    axsize=(2.5, 2.5),
);
```

 So I think it's working as expected (though not handling the de-scan yet).

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+            If True, plot mean diffraction patterns with overlaid origin markers.
+        plot_indices
+            Optional indices to plot. If None, plots all datasets.
+        **plot_kwargs
+            Passed to show_2d.
+
+        Stores
+        ------
+        self.diffraction_origins : np.ndarray
+            Array of shape (n, 2) with integer (row, col) origins.
```
henrygbell, line 156:
```suggestion
            Array of shape (n, 2) with integer (row, col) origins; n = len(datasets).

File: src/quantem/diffraction/maped.py [hunk 3]
```diff
@@ -0,0 +1,892 @@
...
+        scale: float | Sequence[float] | None = None,
+        **plot_kwargs: Any,
+    ) -> MAPED:
+        """
+        Compute dataset summary images.
+
+        Stores
+        ------
+        self.scales : np.ndarray
+            Per-dataset scaling factors (n,).
```
henrygbell, line 76:
```suggestion
            Per-dataset scaling factors (n,); n = len(datasets)
```

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+                1.0 - np.sqrt(dr2) / float(np.mean((H, W))) / float(weight_scale),
+                0.0,
+                1.0,
+            )
+            im_weight = np.sin(im_weight * np.pi / 2.0) ** 2
+
+            shift_rc, G_shift = weighted_cross_correlation_shift(
+                im_ref=G_ref,
+                im=G,
+                weight_real=im_weight * 0.0 + 1.0,
```
henrygbell, line 275: im_weight is not used here. Is that on purpose?

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+                        pass
+
+            dp = getattr(d, "dp_mean", None)
+            if dp is None:
+                arr = np.asarray(d.array)
+                dp_arr = np.mean(arr, axis=(0, 1))
+            else:
+                dp_arr = np.asarray(dp.array if hasattr(dp, "array") else dp)
+
+            arr = np.asarray(d.array)
```
henrygbell, line 114: This is a double conversion, I think one of them is not needed.

henrygbell: Also on line 112, 109.

henrygbell, line 114:
```suggestion
            arr = np.asarray(d)
```

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+
+        merged = np.zeros((Rout, Cout, Hp, Wp), dtype=np.float64)
+
+        dp_local = np.zeros((H, W), dtype=np.float64)
+        dp_pad = np.zeros((Hp, Wp), dtype=np.float64)
+        dp_shifted_tmp = np.zeros((Hp, Wp), dtype=np.float64)
+        num_tmp = np.zeros((Hp, Wp), dtype=np.float64)
+        den_tmp = np.zeros((Hp, Wp), dtype=np.float64)
+
+        for ro in tqdm(range(Rout), desc="Merging (rows)"):
```
henrygbell, line 638: All of these for loops with FFT in the innermost loop (if shift_method = "fourier") will scale poorly for large datasets. I can work on torchifying it!

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+            Array of shape (n, 2) with (row, col) shifts to align diffraction patterns.
+        """
+        if not hasattr(self, "dp_mean"):
+            raise RuntimeError("Run preprocess() first so self.dp_mean exists.")
+        if not hasattr(self, "diffraction_origins"):
+            raise RuntimeError("Run diffraction_origin() first so self.diffraction_origins exists.")
+
+        H, W = np.asarray(self.dp_mean[0]).shape
+
+        w = tukey(H, alpha=2.0 * float(edge_blend) / float(H))[:, None] * tukey(
```
henrygbell, line 247: I think we should clamp the Tukey alphas as it is done in other parts of the code.

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+        self.diffraction_shifts : np.ndarray
+            Array of shape (n, 2) with (row, col) shifts to align diffraction patterns.
+        """
+        if not hasattr(self, "dp_mean"):
+            raise RuntimeError("Run preprocess() first so self.dp_mean exists.")
+        if not hasattr(self, "diffraction_origins"):
+            raise RuntimeError("Run diffraction_origin() first so self.diffraction_origins exists.")
+
+        H, W = np.asarray(self.dp_mean[0]).shape
+
```
henrygbell, line 246:
```suggestion
alpha = min(1.0, 2.0 * float(edge_blend) / float(H))
```

src/quantem/diffraction/maped.py
henrygbell, line 247:
```suggestion
        w = tukey(H, alpha=alpha)[:, None] * tukey(
```

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+        w = tukey(H, alpha=2.0 * float(edge_blend) / float(H))[:, None] * tukey(
+            W, alpha=2.0 * float(edge_blend) / float(W)
```
henrygbell, line 248:
```suggestion
            W, alpha=alpha,
```

src/quantem/diffraction/maped.py
```diff
@@ -0,0 +1,892 @@
...
+            show_2d(
+                [[self.im_bf_merged, self.dp_mean_merged]],
+                title=[["Merged Bright Field", "Merged Mean Diffraction Pattern"]],
+                **plot_kwargs,
+            )
+
+        return dataset_merged
+
+
+def shift_images(
```
henrygbell, line 764: I think the naming of this function is a bit misleading because it shifts a stack of images -- as well as blending them. It's obvious if you glance at the docstring, but I would still change the name to "shift_blend_images" or something like this so there's no confusion. Another way to improve this and make it more general for use elsewhere would be to make flags for return_stack and/or return_blend.

henrygbell (approved):
Testing this code on my Si MAPED dataset, it works very well and required zero tuning.

I do think some minor changes can be made to improve the code, see my comments below.

I think next steps I can take are porting it to torch to speed up the merge_datasets method for large datasets and correcting for de-scan.

henrygbell (approved):

bobleesj:
@henrygbell Thanks Henry! Look forward to testing this code and building an interactive one together. Perhaps could we move forward with this so that we can use this during the exp? 

There are some parts that could be addressed per your comments but I believe we could address this in the following as we are in touch.

bobleesj:
(In other words, please merge if the code is working as intended)

henrygbell:
Since the changes I suggested are very minor, I am merging these changes into nanobeam. Thank you everyone for your comments/testing!

bobleesj:
Since @henrygbell needs to be a part of the org, I believe, no merge right yet. I will merge on behalf as we spoke.

PR #171, opened 2026-02-03
Fix type hints for modify_in_place overloads in Dataset

arthurmccray:
This PR fixes type hints for `Dataset`, specifically to do with the overloading of `Dataset.crop()` and `Dataset.bin()` and `modify_in_place:bool`. 

Previously, `.crop()` -> `Dataset` no matter what, leading to issues: 

image

Now with the fix, everything works as expected. 
image

There are 2 other main types of change in the file: 
- changing `np.isscalar(val)` to `isinstance(val, float|int)` which is better handled by static type checkers 
- ruff reformatted `dataset.py` in a few places due to line lengths--not sure why that hadn't happened previously 

Also, I'm not quite sure why, but `reveal_type` from `typing` was actually working okay even with the old version... and this fix doesn't break it or anything so there's no problems there.

### DIFF
--- a/src/quantem/core/datastructures/dataset.py
+++ b/src/quantem/core/datastructures/dataset.py
@@ -1,5 +1,5 @@
-import os
 import numbers
+import os
 from pathlib import Path
 from typing import Any, Literal, Optional, Self, Union, overload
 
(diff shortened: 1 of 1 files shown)


arthurmccray:
@gvarnavi  let me know if this makes sense to you as i know you've played with static type checkers quite a bit.

gvarnavi (approved):
lgtm, and makes my type-hinter happy!

PR #177, opened 2026-02-20
Add polar transform and pair distribution function analysis

ehrhardtkm:
### What does this PR do?

Implementation of polar coordinate transforms and computing pair distribution functions (PDFs) for 4DSTEM diffraction data. The PDF relies on a polar transformed data structure, which has been implemented in `Polar4dstem`. The code will be updated in the future with other useful tools, like beam stop masking and elliptical corrections.

The workflow takes a 4DSTEM dataset through the following steps:

1. **Origin finding** — Automatically locate the center of each diffraction pattern by minimizing angular intensity variation (`auto_origin_id`)
2. **Polar transform** — Convert diffraction patterns from Cartesian to polar coordinates using torch `grid_sample` for GPU acceleration (`dataset4dstem_polar_transform`)
3. **PDF calculation** — Compute the reduced PDF G(r) from the azimuthally averaged intensity, including background fitting, structure factor extraction, and a windowed sine transform (`calculate_Gr`)
4. **Density estimation** *(optional)* — Iteratively estimate atomic number density and dampen unphysical low-r oscillations using the Yoshimoto-Omote method (`estimate_density`, `calculate_gr`)

#### New files

- `polar4dstem.py` — `Polar4dstem` dataset class for polar-coordinate 4D-STEM data, plus the origin finding and polar transform functions
- `polar.py` — Contains `PairDistributionFunction` class. Handles the full PDF pipeline: radial averaging, background fitting, structure factor calculation, windowed sine transform to G(r), iterative density estimation, and conversion to g(r)
- `tests/diffraction/test_polar.py` — Tests covering construction, transforms, background fitting, PDF calculation, density estimation, and end-to-end workflows

#### Modified files

- `dataset4dstem.py` — Added `polar_transform()` convenience method on `Dataset4dstem`
- `__init__.py` files — Exposed new classes and modules

A companion tutorial notebook ([PDFanalysis_simulatedTa.ipynb](https://github.com/user-attachments/files/25432678/PDFanalysis_simulatedTa.ipynb)) demonstrating an example workflow on simulated amorphous Ta data is available.

---

Provide feedback on the API design and implementation. Key areas for review:

- [ ] `Polar4dstem` class design and its relationship to `Dataset4dstem`
- [ ] PDF pipeline correctness (background subtraction, structure factor normalization, Fourier transform)
- [ ] Ensure the example notebook runs correctly
- [ ] Note any design patterns that were not followed or potential bugs

Checklist items:
- [x] This PR introduces a public-facing change (e.g., API, CLI input/output).
    - [x] Documentation (e.g., tutorials, examples, README) has been updated.
    - [ ] A tracking issue or plan to update documentation exists.
- [ ] This PR affects internal functionality only (no user-facing change).

---

### Reviewer checklist

- [ ] Imported or adapted code has a compatible license


bobleesj:
I'm happy to review this PR since this would naturally extend along `Show4D` widget that you could extend and I was a PDF group for x-ray.

If possible, if you could copy/paste API design here as a comment and attach screenshots (plots, before and after) - these can be helpful as well. API design, as a public facing component, is becoming more important since those are what humans mostly these days.

The attached notebook is always good for the reviewers, but PRs contain global URLs so it's a great place to market/demo to the whole group and the world.

ehrhardtkm:
### Step 1: Find diffraction centers (method on Dataset4dstem)
```python
# Coarse-to-fine origin finding by minimizing angular std
origins = ds.auto_origin_id(
    *, ellipse_params=None,             # (a, b, theta_deg) distortion correction
    num_annular_bins=180,
    radial_min=0.0,                     # pixels
    radial_max=None,                    # pixels (defaults to max radius)
    radial_step=1.0,                    # pixels
    two_fold_rotation_symmetry=False,
    device="cpu",
)
# -> NDArray (scan_y, scan_x, 2) of (row, col) origins in pixels
```

### Step 2: Polar transform (method on Dataset4dstem)
```python
# Cartesian -> polar via torch grid_sample
polar = ds.polar_transform(
    origin_array=None,                  # None | (2,) | (scan_y, scan_x, 2)
    ellipse_params=None,                # (a, b, theta_deg) distortion correction
    num_annular_bins=180,
    radial_min=0.0,                     # pixels
    radial_max=None,                    # pixels
    radial_step=1.0,                    # pixels
    two_fold_rotation_symmetry=False,
    name=None, signal_units=None,
    scan_pos=None,                      # if set, returns a torch.Tensor from a single scan position
    device="cpu",
)
# -> Polar4dstem (or torch.Tensor if scan_pos given)
```

### Step 3: PDF construction
```python
from quantem.diffraction import PairDistributionFunction

pdf = PairDistributionFunction.from_data(
    data,                               # NDArray (2D/4D) | Dataset2d | Dataset3d | Dataset4dstem | Polar4dstem
    *, find_origin=True,
    origin_row=None, origin_col=None,
    ellipse_params=None,                # (a, b, theta_deg) distortion correction
    num_annular_bins=180,
    radial_min=0.0, radial_max=None, radial_step=1.0,
    two_fold_rotation_symmetry=False,
    device="cpu",
)
```

### Step 4: PDF pipeline
```python
# Azimuthal average -> I(k)
pdf.calculate_radial_mean(mask_realspace=None, returnval=False)

# (optional, called automatically by calculate_Gr if needed)
pdf.fit_bg(Ik, kmin, kmax)  # -> (bg, f)

# F(k) + windowed sine transform -> G(r)
# calls calculate_radial_mean + fit_bg internally if not already run
pdf.calculate_Gr(
    k_min_fit=None, k_max_fit=None,       # bg fitting range (1/A)
    k_min_window=None, k_max_window=None,  # Lorch window range; defaults to fit range (1/A)
    k_lowpass=None, k_highpass=None,        # (1/A)
    r_min=0.0, r_max=20.0, r_step=0.02,    # (A)
    mask_realspace=None,
    damp_origin_oscillations=False,         # runs estimate_density if True
    density=None, r_cut=0.8,
    returnval=False,
)  # -> [r, G(r)] if returnval

# (optional) Yoshimoto & Omote iterative density estimation
pdf.estimate_density(density=None, r_cut=0.8, max_iter=40, tol_percent=1e-4)
# -> (rho0, Fk_damped, G_corrected)

# G(r) -> g(r)
pdf.calculate_gr(density=None, r_cut=0.8, set_pdf_positive=False, returnval=False)
# -> [r, g(r)] if returnval
```

### Plotting
```python
pdf.plot_pdf_results(which=("reduced_pdf",), qmin=None, qmax=None,
                     rmin=None, rmax=None, figsize=(6,4), returnfigs=False)
# which: "radial_mean" | "background_fits" | "reduced_sf" |
#        "reduced_pdf" | "pdf" | "oscillation_damping"

# Individual plots (all accept qmin/qmax or rmin/rmax, figsize, returnfig)
pdf.plot_radial_mean()
pdf.plot_background_fits()
pdf.plot_reduced_sf()
pdf.plot_reduced_pdf()
pdf.plot_pdf()
pdf.plot_oscillation_damping()
```

---

### Typical usage
```python
import quantem as em
from quantem.diffraction import PairDistributionFunction

# Load 4DSTEM data
ds = em.core.io.read_4dstem("path/to/data.h5", file_type="arina")

# Separate steps:
origins = ds.auto_origin_id()
polar = ds.polar_transform(origin_array=origins)
pdf = PairDistributionFunction.from_data(polar)

# Convenient one step alternative (auto origin + transform done internally):
# pdf = PairDistributionFunction.from_data(ds)

# PDF analysis
r, Gr = pdf.calculate_Gr(k_min_fit=0.02, k_max_fit=2.0, returnval=True)
r, gr = pdf.calculate_gr(returnval=True)
pdf.plot_pdf_results(which=["reduced_pdf", "pdf"])
```
Updated notebook is available: [PDFanalysis_simulatedTa.ipynb](https://github.com/user-attachments/files/26069071/PDFanalysis_simulatedTa.ipynb)

ehrhardtkm:
The auto origin finding can be evaluated by plotting the polar transform:
￼
Screenshot 2026-03-17 at 2 43 19 PM


When calculating the G(r), the user can either let the I(k) and background fitting run on it’s own behind the scenes by directly calling `rdf.calculate_Gr()` or compute these steps separately. Either way, they will produce the same result, given the same parameters.

Screenshot 2026-03-17 at 2 44 55 PM

Screenshot 2026-03-17 at 2 45 03 PM



Both cells above yield the same output, which can be visualized with the plotting functions.
￼
Screenshot 2026-03-17 at 11 44 49 AM


The small oscillations near 0 of the G(r) can be damped by setting `damp_origin_oscillation=True`. The changes made by the oscillation damping process can be visualized with the `”oscillation_damping”` plotting option. 
￼
Screenshot 2026-03-17 at 12 34 39 PM

src/quantem/diffraction/polar.py
```diff
@@ -136,36 +148,58 @@ def from_data(
...
-            ny, nx = ds4.array.shape[-2:]
-            if origin_row is None:
-                origin_row = (ny - 1) / 2.0
-            if origin_col is None:
-                origin_col = (nx - 1) / 2.0
 
-            polar = ds4.polar_transform(
-                origin_row=origin_row,
-                origin_col=origin_col,
+        # Dataset4dstem input: polar-transform it
```
ehrhardtkm, line 229: Basic constructor workflow is dataset4dstem --> fit origin (if called) --> polar transform. After this point we just work with the polar transformed diffraction data

src/quantem/diffraction/polar.py
```diff
@@ -136,36 +148,58 @@ def from_data(
...
 
         # Dataset3d input: not yet specified how to interpret
         if isinstance(data, Dataset3d):
             raise NotImplementedError(
-                "RadialDistributionFunction.from_data does not yet support Dataset3d inputs."
+                "PairDistributionFunction.from_data does not yet support Dataset3d inputs. "
+                "Please provide a 4D-STEM dataset or a 2D diffraction pattern."
             )
 
         # Numpy array input
```
ehrhardtkm: If np array, load as dataset2d/4d and then fall back through the workflow again to get the polar transformed data

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        if mode == "nearest":
+            x = F.pad(x, (padding, padding), mode="replicate")
+            result = F.conv1d(x, kernel_w)
+        else:
+            result = F.conv1d(x, kernel_w, padding=padding)
+        return result.squeeze(0).squeeze(0)  # reshape to (length)
 
-        Intended to compute radial mean, variance, and normalized variance
-        from self.polar. Not implemented yet.
+    @staticmethod
```
ehrhardtkm: This model is a torchified version of the model first implemented in [py4dstem](https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/polar/polar_analysis.py). It is used to fit the background scattering from the total intensity, which will later be subtracted before computing the PDF.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        """Torch version of the scattering model."""
+        # Add small epsilon to denominators to prevent division by zero during backprop
+        # while still allowing s0/s1 to vary freely
+        eps = 1e-10
+        exp1 = torch.clamp(k2 / (-2.0 * (s0**2 + eps)), min=-100, max=0)
+        exp2 = torch.clamp((k2**2) / (-2.0 * (s1**4 + eps)), min=-100, max=0)
+        # scattering model is monotonic, as is physically expected for backgrounds scattering
+        return c + i0 * torch.exp(exp1) + i1 * torch.exp(exp2)
+
+    def _compute_fit_weights(self, k: torch.Tensor, kmin: float, kmax: float) -> torch.Tensor:
```
ehrhardtkm: Emphasize high k for the model fitting because this region is dominated by the background scattering intensity, so the model will be more accurate when fit to this region instead of overly weighting the low k region

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        i1 = F.softplus(theta[3])
+        s1 = F.softplus(theta[4])
+
+        pred = self._scattering_model_torch(k2, c, i0, s0, i1, s1)
+        residuals = (pred - Ik_norm) ** 2
+        loss = (residuals / (weights**2)).sum()
+        loss.backward()
+        return loss
+
+    def _frequency_filtering(
```
ehrhardtkm: Can be useful for removing unphysical frequencies prior to performing the FT.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        elif k_lowpass is not None and k_lowpass > 0.0:
+            Fk = self._gaussian_filter1d_torch(Fk, sigma=k_lowpass / dk.item(), mode="nearest")
+        elif k_highpass is not None and k_highpass > 0.0:
+            Fk_high = self._gaussian_filter1d_torch(
+                Fk, sigma=k_highpass / dk.item(), mode="nearest"
+            )
+            Fk = Fk - Fk_high
+        return Fk
+
+    def _lorch_window(self, k: torch.Tensor, kmin: float, kmax: float) -> torch.Tensor:
```
ehrhardtkm: This window is used to dampen very low and very high k. Without it, the structure factor deviate from zero, resulting in unphysical ripples that dominate the PDF after the FT. Applying the Lorch window function results in removing a small amount of resolution in the PDF in exchange for removing significant artifacts that can make the PDF otherwise uninterpretable.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+            x == 0,
+            torch.ones_like(x),
+            torch.sin(torch.pi * x) / (torch.pi * x),
+        )
+        lorch = torch.where(inband, sinc_val, torch.zeros_like(k))
+        wk = wk * lorch
+        return wk
 
-    def plot_radial_var_norm(
+    def _compute_alpha_beta(
```
ehrhardtkm: A helper function to perform the density estimation/low-r correction for the G(r). See [this paper ](https://journals.jps.jp/doi/10.7566/JPSJ.91.104602?mobileUi=0)for details.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        -------
+        radial_mean : torch.Tensor or None
+            If `returnval=True`, returns the 1D radial mean intensity (Nk,).
+            Otherwise returns None.
+        """
+        polar_np = self.polar.array  # shape: (scan_y, scan_x, phi, k)
+        scan_y, scan_x, n_phi, n_k = polar_np.shape
+        intensity_sum = torch.zeros(n_k, device=self.device, dtype=torch.float64)
+        n_valid = 0
+        chunk_y = 16  # number of scan_y to process at a time
```
ehrhardtkm: Calculating I(k) in chunks to reduce peak memory usage (can get large for big experimental datasets otherwise)

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+    def fit_bg(
+        self,
+        Ik: torch.Tensor,
+        kmin: float,
+        kmax: float,
+    ) -> tuple[torch.Tensor, torch.Tensor]:
         """
-        Stub for plotting normalized radial variance vs scattering vector.
+        Fit a smooth background B(k) to a radial intensity curve I(k) using
+        PyTorch LBFGS optimizer, with weighting that downweights the low-k
```
ehrhardtkm, line 392: This optimizer was chosen empirically -- it was the one that worked the best through trial and error

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+            s1 = F.softplus(theta[4])
+            # undo normalization
+            c_scaled = c * int_mean
+            i0_scaled = i0 * int_mean
+            i1_scaled = i1 * int_mean
+            # compute bg and the average scattering factor f(k)
+            bg = self._scattering_model_torch(k2, c_scaled, i0_scaled, s0, i1_scaled, s1)
+            f = bg - c_scaled
+        self.bg = bg
+        self.f = f
```
ehrhardtkm, line 506: f is the estimated average atomic scattering factor. We need it to compute the reduced structure factor F(k) later

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        return bg, f
+
+    def calculate_Gr(
         self,
-        k_min: float = 0.05,
-        k_max: float | None = None,
-        k_width: float = 0.25,
+        k_min_fit: float | None = None,
+        k_max_fit: float | None = None,
+        k_min_window: float | None = None,
```
ehrhardtkm, line 513: I've found better success when we can fit the background to a wider k range than the k range we use for the lorch window, so they're split up into two sets of arguments here. Users will have to play with the exact values to use for their data regardless.

bobleesj:
@ehrhardtkm is it ready to be reviewed?

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+
+        Returns
+        -------
+        list[np.ndarray] or None
+        """
+        # clear results from any previous run so stale state doesn't leak
+        self.Fk_damped = None
+        self.reduced_pdf_damped = None
+        self.rho0 = None
+        # this is missing a 2pi term that we add back during the pdf calc later
```
ehrhardtkm, line 592: py4dstem was missing this 2pi term, so that is a difference between this implementation and the one there.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        -------
+        list[np.ndarray] or None
+        """
+        if self._reduced_pdf is None or self._r is None:
+            raise RuntimeError(
+                "Reduced PDF not computed."
+                "Run PairDistributionFunction.calculate_Gr() before calculate_gr()."
+            )
+
+        # Determine density
```
ehrhardtkm, line 732: g(r) needs a number density to be computed. Ideally the user would know the value to use every time, but having a way to estimate it to helpful in the event that they do not but still want a qualitative g(r) for their analysis.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        if density is not None:
+            rho0 = density
+        elif self.rho0 is not None:
+            rho0 = self.rho0
+            print(f"  Using estimated rho0 = {rho0:.6f} atoms/A^3", flush=True)
+        else:
+            # the oscillation correction simultaneously produces a density estimate
+            # if the user didn't run damping in calculate_Gr, we can still run the density estimation without using the corrected Fk/G(r)
+            density_est = self.estimate_density(
```
ehrhardtkm, line 741: This function does both the density estimation and dampens ripple artifacts at low-r at the same time. This is because the ideal low-r region that we are correcting to is proportional to the number density as r approaches 0.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+                "Run PairDistributionFunction.calculate_Gr() before estimate_density()."
+            )
+
+        k = self._to_torch(np.asarray(self.qq))
+        dk = k[1] - k[0]
+        k_fit_mask = (k >= self.kmin_fit) & (k <= self.kmax_window)
+        k_fit = k[k_fit_mask]
+        ka, ra = torch.meshgrid(k, self._r, indexing="ij")
+
+        # r_cut sets the minimum r for the peak search used to determine the correction interval
```
ehrhardtkm, line 829: The correction should only be applied to the low-r region. This region will depend on the PDF and where the peaks are, so we find the major peak (should be the first peak unless there is a lot of dynamical scattering, in which case there are already issues with data quality and they probably shouldn't use this correction in the first place) and then use the local minimum to the left of that peak as the r_max for our iterative correction

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+                rmin = float(r_left[mins_indices[-1]])
+            else:
+                rmin = float(r_left[torch.argmin(G_left)])
+        # Restrict r to [0, rmin] for the correction
+        r_mask = (self._r >= 0.0) & (self._r <= rmin)
+        r_short = self._r[r_mask]
+        k_fit_scaled = k_fit * 2 * torch.pi
+        k2d_fit, r2d_fit = torch.meshgrid(k_fit_scaled, r_short, indexing="ij")
+
+        # Iterative refinement of rho0 and S(k)
```
ehrhardtkm, line 857: The iteractive correction is actually applied to the structure factor S(k), then we use the corrected S(k) to recompute the PDF and see if it's better

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+        u = (a / b) * r * torch.cos(alpha)
+        v_prime = r * torch.sin(alpha)
+        cos_t = torch.cos(theta)
+        sin_t = torch.sin(theta)
+        x = u * cos_t - v_prime * sin_t
+        y = u * sin_t + v_prime * cos_t
+    return x, y
+
+
+def _build_candidate_grids(
```
ehrhardtkm: When running auto_origin_id(), we trial many different pixels as the origin. Pre-computing the sampling grids for each pixel that we test as origin is much more efficient. The sampling grids of two different pixels differ by an offset, which is applied here to each grid as a batch.

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -161,42 +519,120 @@ def dataset4dstem_polar_transform(
...
+    elif origin_array.shape == (scan_y, scan_x, 2):
+        origins = origin_array
+    else:
+        raise ValueError(
+            f" Got {origin_array.shape}. "
+            "origin_array must have shape None, (2,) or (scan_y, scan_x, 2)."
+        )
+
+    # If scan_pos is provided, compute polar transform only for that position
+    if scan_pos is not None:
```
ehrhardtkm: Probably not a common use case, but if a user only wants the origin coords of one scan position in a 4DSTEM scan, can be accomplished here

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -161,42 +519,120 @@ def dataset4dstem_polar_transform(
...
+            grid,
+            mode="bilinear",
+            padding_mode="zeros",
+            align_corners=True,
+        )
+        return polar2d.squeeze(0).squeeze(0)  # (n_phi, n_r)
+
+    # Use the global minimum safe radius across all origins so every scan
+    # position maps to the same-size polar grid (required for a uniform 4D output)
+    if radial_max is None:
```
ehrhardtkm: To allow consistency in the size of each polar transformed scan position, we use the largest radius possible that is still safe for all scan positions. Otherwise, downstream analysis of data with different sizes gets tricky

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+        global_row,
+        global_col,
+        local_margin,
+        ny,
+        nx,
+        x_norm_scale,
+        y_norm_scale,
+        device,
+    )
+    pbar = tqdm(total=total_positions, desc="Finding origin for each scan position")
```
ehrhardtkm: This is the slow part (~20 min), so adding a progress bar here.

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+    grid_sample expects x_norm = 2*x/(W-1) - 1, y_norm = 2*y/(H-1) - 1,
+    stacked as (..., 2) in [x, y] order.
+    """
+    x_norm = 2.0 * coords_x / (width - 1) - 1.0
+    y_norm = 2.0 * coords_y / (height - 1) - 1.0
+    return torch.stack([x_norm, y_norm], dim=-1)
+
+
+def _polar_to_cartesian_offsets(
+    phi: torch.Tensor,
```
ehrhardtkm: These are the cartesian offsets to apply to a polar transformed sampling grid relative to (0,0) in the original data. The ellipse correction was implemented by Colin but the code to actually compute the elliptical correction terms will probably be in a future PR.

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: Since auto_origin_id() trials many different pixels as the origin, batch pre-computing the sampling grids is much more efficient. We just apply the pixel offsets computed elsewhere in this file so the grid corresponds with each pixel that we are treating as the origin

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+        radial_bins = torch.tensor([radial_min], dtype=torch.float32, device=device)
+    phi_range = torch.pi if two_fold_rotation_symmetry else 2.0 * torch.pi
+    # Drop the last endpoint because 0 and 2pi (or pi) are the same angle
+    phi_bins = torch.linspace(
+        0.0, phi_range, num_annular_bins + 1, dtype=torch.float32, device=device
+    )[:-1]
+    phi_grid, r_grid = torch.meshgrid(phi_bins, radial_bins, indexing="ij")
+    # Compute offsets relative to origin (0,0) so they can be reused
+    # for any candidate origin by simple translation
+    offset_x, offset_y = _polar_to_cartesian_offsets(phi_grid, r_grid, ellipse_params, device)
```
ehrhardtkm: Goal here is to compute offsets once to avoid repetitive computation

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+            (nx - 1) - (com_col + global_margin),
+        )
+    )
+    if radial_max is not None:
+        safe_rmax = min(safe_rmax, float(radial_max))
+    if safe_rmax <= radial_min:
+        safe_rmax = radial_min + radial_step
+    # use very coarse binning because asymmetry is still captured at
+    # low angular resolution and is significantly faster
+    search_n_phi = 36
```
ehrhardtkm: Not sure if search_n_phi and coarse_step are the best possible values, but this will probably also vary between datasets...

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+        x_norm_scale,
+        y_norm_scale,
+        device,
+    )
+    fine_scores = _angular_std_scores(mean_dp_batch, fine_grids, min_r_idx, max_r_idx)
+    best_fine_idx = fine_scores.argmin().item()
+    global_row = int(fine_row_flat[best_fine_idx].item())
+    global_col = int(fine_col_flat[best_fine_idx].item())
+    # Get center for each scan pos by fine search around global center
+    # Assuming that the center doesn't shift more than 10 pixels across the scan
```
ehrhardtkm: Similarly here, not sure if this is an appropriately large value. Maybe go larger?

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: This is the slow part (~20 min), so adding a progress bar here

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -161,42 +519,120 @@ def dataset4dstem_polar_transform(
...
+        origins[...] = origin_array
+    elif origin_array.shape == (scan_y, scan_x, 2):
+        origins = origin_array
+    else:
+        raise ValueError(
+            f" Got {origin_array.shape}. "
+            "origin_array must have shape None, (2,) or (scan_y, scan_x, 2)."
+        )
+
+    # If scan_pos is provided, compute polar transform only for that position
```
ehrhardtkm: Probably an uncommon use case, but if the user only cares about a certain scan position, they can just find the origin of that position here

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: The polar transformed data should have the same shape across all scan positions, otherwise the downstream analysis gets tricky. To ensure this, it finds the largest max radius that is still safe for all scan positions

src/quantem/diffraction/polar.py
```diff
@@ -86,47 +106,39 @@ def from_data(
...
+            defaults to the central pixel of the diffraction pattern.
         Other parameters
             Passed through to Dataset4dstem.polar_transform when needed.
         """
         # Polar input: use directly
         if isinstance(data, Polar4dstem):
             polar = data
-            return cls(polar=polar, input_data=data, _token=cls._token)
-
-        # Dataset4dstem input: polar-transform it
```
ehrhardtkm, line 100: Basic constructor workflow is dataset4dstem --> find origin (if not provided) --> polar transform. If the data is provided as a dataset2d instance, it gets wrapped into a 1x1 dataset4dstem and fed through this pipeline. If data is provided as a numpy array, it is also used to create a dataset4dstem instance before being fed through.

src/quantem/diffraction/polar.py
```diff
@@ -136,36 +148,58 @@ def from_data(
...
                 radial_step=radial_step,
                 two_fold_rotation_symmetry=two_fold_rotation_symmetry,
+                device=device,
             )
-            return cls(polar=polar, input_data=data, _token=cls._token)
+            return cls(polar=polar, input_data=data, device=device, _token=cls._token)
 
         # Dataset3d input: not yet specified how to interpret
         if isinstance(data, Dataset3d):
             raise NotImplementedError(
```
ehrhardtkm: Not sure what Dataset3d will look like yet, leaving it not implemented for now

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+            x = F.pad(x, (padding, padding), mode="replicate")
+            result = F.conv1d(x, kernel_w)
+        else:
+            result = F.conv1d(x, kernel_w, padding=padding)
+        return result.squeeze(0).squeeze(0)  # reshape to (length)
 
-        Intended to compute radial mean, variance, and normalized variance
-        from self.polar. Not implemented yet.
+    @staticmethod
+    def _scattering_model_torch(
```
ehrhardtkm: This model is used to fit the background scattering signal. This is a torchified version of the model implemented originally in [py4dstem.](https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/polar/polar_analysis.py)

src/quantem/diffraction/polar.py
ehrhardtkm: Emphasizing high k over low k is important because background scattering dominates this high k region and we also don't want the fit to be overly fit to the direct beam, which can otherwise strongly shift the model fit because of its large intensity. Similar weighting is performed in py4dstem.

src/quantem/diffraction/polar.py
ehrhardtkm: Frequency filtering can be helpful to remove unphysical frequencies from data. Bad low frequencies can come from poor background subtraction, while bad high frequencies are often due to noise

src/quantem/diffraction/polar.py
ehrhardtkm: Lorch windows are a typical way of windowing the structure factor. Without it, the structure factor may deviate from 0 at the small and large extremes, which will cause significiant ripples/artifacts in PDF that are unphysical and can make the PDF difficult or impossible to interpret. This is an important step of the PDF reconstruction process.

src/quantem/diffraction/polar.py
ehrhardtkm: This is a helper function for the density estimation/artifact damping in estimate_density(). See [Yoshimoto and Omote](https://journals.jps.jp/doi/10.7566/JPSJ.91.104602?mobileUi=0) for details.

src/quantem/diffraction/polar.py
ehrhardtkm: Computing in chunks to avoid large peak memory. My laptop struggled with processing experimental datasets without this

src/quantem/diffraction/polar.py
ehrhardtkm, line 392: Chose the optimizer empirically by seeing what matched scipy.curve_fit() best through trial and error -- don't have a precise technical reason for why this performed best

src/quantem/diffraction/polar.py
ehrhardtkm, line 506: f is the estimated average atomic scattering form factor. We need it later on to compute F(k), the reduced structure factor

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+
+    def calculate_Gr(
         self,
-        k_min: float = 0.05,
-        k_max: float | None = None,
-        k_width: float = 0.25,
+        k_min_fit: float | None = None,
+        k_max_fit: float | None = None,
+        k_min_window: float | None = None,
+        k_max_window: float | None = None,
```
ehrhardtkm, line 514: I've had better success when I use a wider k-range for the fit and a narrower k-range for the structure factor window. Giving users the option to select different ranges here, but the window defaults to the fit window if not provided

src/quantem/diffraction/polar.py
ehrhardtkm, line 592: This 2pi term was missing from the implementation in py4dstem

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        self.Ik = Ik
+        self.bg = bg
+        self.Fk = Fk
+        self.Fk_masked = Fk_win
+        self._r = r
+        self._reduced_pdf = reduced_pdf
+
+        # optionally damped unphysical oscillations near the origin by iteratively estimating density and correcting F(k)
+        if damp_origin_oscillations:
+            density_est = self.estimate_density(
```
ehrhardtkm, line 669: This estimate_density() function calculates the approx. density and applies a correction to reduce the unphysical oscillations as low r simultaneously. This is because the correction is strictly in the low-r region, and the ideal slope of this region is proportional to the number density.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+            )
+
+        k = self._to_torch(np.asarray(self.qq))
+        dk = k[1] - k[0]
+        k_fit_mask = (k >= self.kmin_fit) & (k <= self.kmax_window)
+        k_fit = k[k_fit_mask]
+        ka, ra = torch.meshgrid(k, self._r, indexing="ij")
+
+        # r_cut sets the minimum r for the peak search used to determine the correction interval
+        mask_search = self._r >= r_cut
```
ehrhardtkm, line 830: This correction should only be applied to the low-r region, but the exact range will depend on the data and where the peaks are. Here, it finds the major peak, which is almost always the first peak (unless there is significant dynamical scattering, which means that there is already a data quality issue and they probably shouldn't bother with this correction anyway), and then uses the local min to the left of this peak as the r_max for the range used in the correction.

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
...
+        r_mask = (self._r >= 0.0) & (self._r <= rmin)
+        r_short = self._r[r_mask]
+        k_fit_scaled = k_fit * 2 * torch.pi
+        k2d_fit, r2d_fit = torch.meshgrid(k_fit_scaled, r_short, indexing="ij")
+
+        # Iterative refinement of rho0 and S(k)
+        fixed_density = density is not None
+        rho0 = density if fixed_density else 0.0
+        rho0_prev = None
+        Sk_cor = self.Sk.clone()
```
ehrhardtkm, line 861: The correction is actually applied to the structure factor S(k), which is then used to recompute G(r)

ehrhardtkm (commented):
Adding in-line comments for added clarity

bobleesj:
@ehrhardtkm Thanks Karen! I will review this most likely Friday after the two-day experiment. Look forward to testing this!

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -1,7 +1,12 @@
+from __future__ import annotations
```
bobleesj: don't think we need `Any`  (use native class as much as we can) 

`TYPE_CHECKING` and `annotation` also may not need if we use `Self` for class return

src/quantem/core/datastructures/dataset4dstem.py
```diff
@@ -7,8 +7,12 @@
 
 from quantem.core.datastructures.dataset2d import Dataset2d
 from quantem.core.datastructures.dataset4d import Dataset4d
-from quantem.core.datastructures.polar4dstem import dataset4dstem_polar_transform
-
+from quantem.core.datastructures.polar4dstem import (
+    auto_origin_id as _auto_origin_id,
```
bobleesj: `_` is necessary?

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -62,7 +67,10 @@ def from_array(
     ) -> "Polar4dstem":
```
bobleesj: Self https://peps.python.org/pep-0673/

bobleesj: this pacakge 3.11 so it's just a few prompts away.

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
         return int(self.array.shape[3])
 
 
-def _precompute_polar_coords(
+def _to_numpy(tensor: torch.Tensor) -> NDArray:
```
bobleesj: not urgent - but seems like a util function that can be used outside of this module

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+    """Convert torch tensor to numpy array."""
+    return tensor.detach().cpu().numpy()
+
+
+def _normalize_coords_for_grid_sample(
```
bobleesj: does polar coord use "x" "y" or row or col? the moment we introduce "x" and "y" coordinates become ambiguous across the package described here https://github.com/ophusgroup/dev?tab=readme-ov-file#d4-how-do-i-represent-coordinates-in-numpy-matplotlib-and-quantem 

want to go slow on this

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -91,68 +99,418 @@ def n_r(self) -> int:
...
+    return torch.stack([x_norm, y_norm], dim=-1)
+
+
+def _polar_to_cartesian_offsets(
+    phi: torch.Tensor,
+    r: torch.Tensor,
+    ellipse_params: tuple[float, float, float] | None,
+    device: str = "cpu",
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Convert polar (phi, r) grids to Cartesian (x, y) offsets from the origin,
```
bobleesj: okay, we are using classic polar and cartesian (x,y), this is, obviously, introducing a new convention and x and y might be the correct way to represent this - @ehrhardtkm please let me know accouting the comment above.

src/quantem/core/datastructures/polar4dstem.py
bobleesj: a bit lost - x,y, row, col, ny?, nx... need propert docstring on this - what are they?

bobleesj: private functions generally don't need docstrings since it's meant to be a simple function that "aid" other main functions but I feel like this this a whole function and vectorized so can borrow help

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -161,42 +519,120 @@ def dataset4dstem_polar_transform(
...
+            origin_col=c0,
+            ellipse_params=ellipse_params,
+            num_annular_bins=num_annular_bins,
+            radial_min=radial_min,
+            radial_max=radial_max,
+            radial_step=radial_step,
+            two_fold_rotation_symmetry=two_fold_rotation_symmetry,
+            device=device,
+        )
+        dp_batch = dp.unsqueeze(0).unsqueeze(0)  # (1, 1, ny, nx)
```
bobleesj: either `none` or `unsqueeze`  - usually `none` used

src/quantem/core/datastructures/polar4dstem.py
```diff
@@ -161,42 +519,120 @@ def dataset4dstem_polar_transform(
...
+    scan_pos: tuple[int, int] | None = None,
+    device: str = "cpu",
+) -> Polar4dstem | torch.Tensor:
     if self.array.ndim != 4:
-        raise ValueError("polar_transform requires a 4D-STEM dataset (ndim=4).")
+        raise ValueError(
+            f"Found array with shape: {self.array.shape}. "
+            "polar_transform requires a 4D-STEM dataset (ndim=4)."
+        )
     scan_y, scan_x, ny, nx = self.array.shape
```
bobleesj: y and x? y an x? same comment from the group standard

src/quantem/diffraction/polar.py
```diff
@@ -31,40 +47,42 @@ def __init__(
...
-
-        self.Sk: NDArray | None = None
-        self.fk: NDArray | None = None
-        self.bg: NDArray | None = None
-        self.offset: float | None = None
-        self.Sk_mask: NDArray | None = None
+        self._r: torch.Tensor | None = None
+        self._reduced_pdf: torch.Tensor | None = None
+        self._pdf: torch.Tensor | None = None
+        self.Ik: torch.Tensor | None = None
```
bobleesj, line 139: docstring? Users imports the class and read and see these variables without context or equation. I learned PDF but forgot PDF after a year so I can't learn from this function/class. If needed, add mathematical functions that is rendred in proper API doc.

src/quantem/diffraction/polar.py
```diff
@@ -73,9 +91,11 @@ def from_data(
         radial_max: float | None = None,
         radial_step: float = 1.0,
         two_fold_rotation_symmetry: bool = False,
-    ) -> "RadialDistributionFunction":
+        device: str = "cpu",
+    ):
         """
-        Create a RadialDistributionFunction object from various input types.
+         -> "PairDistributionFunction"
```
bobleesj: Self

src/quantem/diffraction/polar.py
```diff
@@ -73,9 +91,11 @@ def from_data(
...
+        Create a PairDistributionFunction object from various input types.
 
         Parameters
```
bobleesj, line 170: is this numpy api style?

src/quantem/diffraction/polar.py
```diff
@@ -31,40 +47,42 @@ def __init__(
...
+        self.Fk_mask: torch.Tensor | None = None
+        self.Fk_damped: torch.Tensor | None = None
+        self.reduced_pdf_damped: torch.Tensor | None = None
+        self.rho0: float | None = None
 
     # ------------------------------------------------------------------
     # Constructors
     # ------------------------------------------------------------------
     @classmethod
     def from_data(
```
bobleesj, line 153: I'm a bit confused - I thought PDF takes in 4DSTEM dataset and then finds the mask and calculate those 1D plots, why does `data` support a lot more than those types?

src/quantem/diffraction/polar.py
```diff
@@ -31,40 +47,42 @@ def __init__(
...
         cls,
-        data: Union[NDArray, Dataset2d, Dataset3d, Dataset4dstem, Polar4dstem],
+        data: NDArray | Dataset2d | Dataset3d | Dataset4dstem | Polar4dstem,
```
bobleesj: docstring

src/quantem/diffraction/polar.py
```diff
@@ -86,47 +106,39 @@ def from_data(
...
+        # Dataset2d input: wrap as a trivial 4D-STEM (1x1 scan) and fall through
         if isinstance(data, Dataset2d):
             arr2d = data.array
             if arr2d.ndim != 2:
-                raise ValueError("Dataset2d for RDF must be 2D.")
+                raise ValueError(
+                    f"Found array with shape: {arr2d.shape}. "
+                    "Dataset2d for PairDistributionFunction must be 2D."
+                )
             arr4 = arr2d[None, None, ...]  # (1, 1, ky, kx)
```
bobleesj: row col conventions

src/quantem/diffraction/polar.py
```diff
@@ -86,47 +106,39 @@ def from_data(
...
-                radial_min=radial_min,
-                radial_max=radial_max,
-                radial_step=radial_step,
-                two_fold_rotation_symmetry=two_fold_rotation_symmetry,
-            )
-            return cls(polar=polar, input_data=data, _token=cls._token)
+            return cls(polar=polar, input_data=data, device=device, _token=cls._token)
 
-        # Dataset2d input: wrap as a trivial 4D-STEM (1x1 scan) then polar-transform
+        # Dataset2d input: wrap as a trivial 4D-STEM (1x1 scan) and fall through
```
bobleesj, line 204: why 1x1 scan?

src/quantem/diffraction/polar.py
bobleesj: cool -

src/quantem/diffraction/polar.py
```diff
@@ -136,36 +148,58 @@ def from_data(
...
+                    num_annular_bins=num_annular_bins,
+                    radial_min=radial_min,
+                    radial_max=radial_max,
+                    radial_step=radial_step,
+                    two_fold_rotation_symmetry=two_fold_rotation_symmetry,
+                    device=device,
+                )
+            else:
+                if origin_row is None:
+                    origin_row = (ny - 1) / 2.0
```
bobleesj: ny nx

src/quantem/diffraction/polar.py
bobleesj: why dataset3d needed?

src/quantem/diffraction/polar.py
```diff
@@ -204,123 +242,1146 @@ def qq(self) -> Any:
         """
         # Polar4dstem dims: (scan_y, scan_x, phi, r)
```
bobleesj: convention

bobleesj (commented):
@ehrhardtkm I’m having a hard time following the flow of the PDF code, as many functions are scattered around and conventions like row/col vs. x/y and nx/ny are not used consistently, and some key docstrings are missing where they would be helpful. 

For this iteration, I tried to understand the code without executing the notebook but couldn’t form a clear, consistent picture, so I’d recommend avoiding adding placeholder features that aren’t implemented yet, and moving general-purpose functions like the Gaussian filter into a utils module so the PDF code itself tells a more coherent, uninterrupted story. You've mentioned this code https://github.com/py4dstem/py4DSTEM/blob/dev/py4DSTEM/process/polar/polar_analysis.py and I generally like how compact and uninterrupted they are with clear docstrings and mathematical key functions, where mathematical variable sare defined. I believe this PR iteration will take many reviews, so please feel free to suggest a specific region that you want me to review if time isn't the main constraint here. 

Just to add one more comment.  If a code block is used once, don't create a function. Including private ones. Another comment is that - does it have to be 2,000 lines of code? Why not 1,000? is it the absolute minimum necessary to deliver this PDF feature? While clarity is maintained, we always want to do minimum.

src/quantem/diffraction/polar.py
ehrhardtkm: This is useful if people want to perform this analysis on a single diffraction pattern, whether that's from TEM (SAED) or an averaged diffraction pattern over a sample area. So I expect this to be a real use case that users will want.

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: Thanks for directing to the group guidelines. I've changed the coordinate conventions used across the PR to be consistent so that the pre-polar transform data is indexed using `row`\col as the guidelines state. Polar coordinates use `phi`/`r_pix` instead of x/y now to avoid ambiguity. Plotting and pytorch API conflicts are the only exceptions, I think. (torch's `grid_sample` needs `col`/`row`, but this is noted in the comments where it occurs). Please let me know if you run into any parts that I missed.

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: I chose to stick with row/col for the cartesian offsets because they still use the same coordinate system

src/quantem/core/datastructures/polar4dstem.py
ehrhardtkm: Added a docstring for clarity. The confusion with x, y, etc. should also be resolved by the coordination convention rename described in comments above. Essentially, there's three row/col pairs in this function: the polar sampling offsets, the center of the origin search window, and the DP dimensions

src/quantem/diffraction/polar.py
ehrhardtkm: Should be renamed to n_row/n_col throughout now

src/quantem/diffraction/polar.py
ehrhardtkm: Removed Dataset3dstem for now and can add it back later if needed

ehrhardtkm:
Sorry for the delay. I did some refactoring to try to make the flow easier to follow, and also cleared up the x/y vs row/col conventions throughout the PR. Hopefully, that addresses some of the concerns with clarity. Specifically, I've refactored the code to split it up along the two major steps: `polar_transform.py` handles the transformation to polar coordinates and finding the center of the DPs, while `polar.py`  contains the code for analysis of the transformed data. Hopefully this makes the flow easier to follow as a sequence instead of scattering functinons.

On minimizing length: I went through and removed some one-time helpers and replaced some others with some general utility functions in the repo. It's still a similar overall length, but I'm not quite sure what else I'd cut while keeping the current flexibility during PDF reconstruction. If you can flag any blocks that look redundant or unnecessary, I will definitely take a look! 

A side note is that almost identical gaussian filter helper functions are in `tomography/utils.py` but I needed a 1d version, so I added that there. It might be a larger discussion on whether these utils should be somewhere more general than explicitly under tomo.

bobleesj:
@ehrhardtkm Thanks Karen - will review probalby tm!

src/quantem/diffraction/polar.py
bobleesj: thanks!

src/quantem/diffraction/polar.py
bobleesj: done

src/quantem/diffraction/polar.py
```diff
@@ -1,70 +1,127 @@
...
+        ⟨f(k)⟩². Set by ``fit_bg``.
+    Sk : torch.Tensor or None
+        Structure factor S(k) = 1 + [I(k) − B(k)] / f(k). Set by ``calculate_Gr``.
+    Fk : torch.Tensor or None
+        Reduced structure factor F(k) = 2π · k · [S(k) − 1]. The 2π factor is
+        explicitly included (py4dstem omits it). Set by ``calculate_Gr``.
+    Fk_mask : torch.Tensor or None
+        Window applied to F(k) before the sine transform, combining a bandpass
+        and a Lorch taper. Set by ``calculate_Gr``.
+    Fk_damped : torch.Tensor or None
```
bobleesj, line 77: i will just assume that these are math symbols that are used in practice/paper/lit

bobleesj: for all symbols

src/quantem/diffraction/polar.py
```diff
@@ -1,70 +1,127 @@
...
+
+    r : NDArray or None
+        Real-space radial axis in Å.
+    reduced_pdf : NDArray or None
+        Reduced pair distribution function G(r), obtained by windowed sine
+        transform of F(k):
+        G(r) = (2/π) ∫ F(k) · sin(2π · k · r) dk.
+    pdf : NDArray or None
+        Pair distribution function g(r) = 1 + G(r) / (4π · r · ρ₀).
     """
```
bobleesj, line 114: Example section needed in API docstring on how to use the class at least for beginners. 

Image

it's already provided to let's have sth simliar: 

ref: https://github.com/ophusgroup/dev#d1-how-do-i-write-effective-docstrings

src/quantem/diffraction/polar.py
bobleesj: @ehrhardtkm if you've addressed some of the comments by reviwer, pls either give me a thumbs-up or write "done" etc. so that 

Image

src/quantem/tomography/utils.py
```diff
@@ -157,6 +157,17 @@ def gaussian_filter_2d(
     return img.squeeze_(0).squeeze_(0)  # Make 2D again
 
 
+def gaussian_filter_1d(
```
bobleesj: prob in https://github.com/electronmicroscopy/quantem/blob/dev/src/quantem/core/utils/imaging_utils.py
? 

tomo util should be only used by tomo people in general, so later we can move some tomo utils into general imaging uttils but later refactoring

src/quantem/diffraction/polar.py
bobleesj: okay, this "why 1x1x scan" shall be added either in the dcostring or the comment clearly for future devs

bobleesj: ref

Image

src/quantem/diffraction/polar.py
bobleesj: just to confirm - is this "general workflow" provided in the docstring - perhaps at the class level? 

ref: https://github.com/electronmicroscopy/quantem/pull/204/changes#r3128048826

sorry - it's 2k lines of code so a bit hard to navigate for me

bobleesj: docstring needs to make a very good connection b/w dataset4dstem and polar4dstem datasets so that we know why 2 things exist.

src/quantem/diffraction/polar.py
```diff
@@ -136,191 +199,1154 @@ def from_data(
...
-                radial_step=radial_step,
-                two_fold_rotation_symmetry=two_fold_rotation_symmetry,
-            )
-        else:
-            raise ValueError(
-                "RadialDistributionFunction.from_data only supports 2D or 4D arrays."
-            )
+            return cls(polar=polar, input_data=data, device=device, _token=cls._token)
+
+        raise TypeError(
```
bobleesj, line 265: very good error msg

src/quantem/diffraction/polar.py
```diff
@@ -1,70 +1,127 @@
...
+    Fk_damped : torch.Tensor or None
+        F(k) after iterative low-r oscillation damping, set by ``estimate_density``.
+    reduced_pdf_damped : torch.Tensor or None
+        G(r) recomputed from the damped F(k), set by ``estimate_density``.
+    rho0 : float or None
+        Estimated atomic number density ρ₀ (atoms/Å³), set by ``estimate_density``.
+
+    Exposed as read-only numpy-valued properties:
+
+    r : NDArray or None
```
bobleesj, line 86: i think `r` makes sense here, better than, say, `real_space_radial`, etc. sth like that and that there is the "grouth truth" as to what these symbols represent compared to general variable names human-made.

src/quantem/diffraction/polar.py
bobleesj: since most people will use using the notebook that you've provided with the explicit steps includinig to use `dataset4dsteem` or `dataset2d` and the API reference,

why can't we just use `dataset2d` and `dataset4dstem` only? I think it's better to have this class focus on science/math rather than handling different types like handing `numpy` and `torch`, `cupy`, etc.

bobleesj: we want our class to be flexible in the future - most of my 4dstem data will be either in torch or cupy (never nump) so handling all different data types isn't the job of `polar.py`.

src/quantem/diffraction/polar.py
bobleesj: this comment addreses "why' - let's add a quick statement into the docstring since it addresses "why LBFGS".

bobleesj: same comment

src/quantem/diffraction/polar.py
bobleesj: yup, the addresss "why this code exists" so a quick sentence in the docstring

bobleesj: same comment

src/quantem/diffraction/polar.py
```diff
@@ -136,191 +199,1154 @@ def from_data(
...
-        plot_reduced_pdf: bool = True,
-        plot_pdf: bool = False,
-        figsize: tuple[float, float] = (8, 4),
-        maxfev: int | None = None,
+        r_cut: float = 0.8,
         returnval: bool = False,
-        returnfig: bool = False,
+    ) -> list[NDArray] | None:
+        """
+        Calculate the reduced pair distribution function G(r) from a 4D-STEM dataset.
```
bobleesj, line 527: Great great workflow

src/quantem/diffraction/polar_transform.py
```diff
@@ -0,0 +1,621 @@
...
+            "origin_array must have shape None, (2,) or (scan_row, scan_col, 2)."
+        )
+
+    # If scan_pos is provided, compute polar transform only for that position
+    if scan_pos is not None:
+        i_row, i_col = scan_pos
+        dp = torch.from_numpy(data.array[i_row, i_col].astype(np.float32)).to(device)
+        r0 = float(origins[i_row, i_col, 0])
+        c0 = float(origins[i_row, i_col, 1])
+        grid, phi_bins, radial_bins, radial_max_eff = _precompute_polar_coords(
```
bobleesj: we are repeaitng 

```
_precompute_polar_coords(
            n_row=n_row,
            n_col=n_col,
            origin_row=r0,
            origin_col=c0,
            ellipse_params=ellipse_params,
            num_annular_bins=num_annular_bins,
            radial_min=radial_min,
            radial_max=radial_max,
            radial_step=radial_step,
            two_fold_rotation_symmetry=two_fold_rotation_symmetry,
            device=device,
        )

_precompute_polar_coords(
            n_row=n_row,
            n_col=n_col,
            origin_row=r0,
            origin_col=c0,
            ellipse_params=ellipse_params,
            num_annular_bins=num_annular_bins,
            radial_min=radial_min,
            radial_max=radial_max,
            radial_step=radial_step,
            two_fold_rotation_symmetry=two_fold_rotation_symmetry,
            device=device,
        )

_precompute_polar_coords(
            n_row=n_row,
            n_col=n_col,
            origin_row=r0,
            origin_col=c0,
            ellipse_params=ellipse_params,
            num_annular_bins=num_annular_bins,
            radial_min=radial_min,
            radial_max=radial_max,
            radial_step=radial_step,
            two_fold_rotation_symmetry=two_fold_rotation_symmetry,
            device=device,
        )
```

in this code block, not so clean. any cleaner approach possible?

src/quantem/diffraction/polar.py
```diff
@@ -136,191 +199,1154 @@ def from_data(
...
     ):
         """
-        Stub for plotting background fit vs radial mean intensity.
+        Plotting background fit vs radial mean intensity.
         """
-        raise NotImplementedError("plot_background_fits is not implemented yet.")
+        if self.Ik is None or self.bg is None:
+            raise RuntimeError(
+                "Radial mean intensity or background has not been calculated yet."
+                "Run PairDistributionFunction.calculate_Gr() or both calculate_radial_mean() and calculate_background() before plotting."
```
bobleesj, line 1015: great error msg

src/quantem/diffraction/polar_transform.py
```diff
@@ -0,0 +1,621 @@
...
+            best_idx = scores.argmin().item()
+            origin_array[row_pos, col_pos, 0] = local_rf[best_idx].item()
+            origin_array[row_pos, col_pos, 1] = local_cf[best_idx].item()
+            pbar.update(1)
+
+    pbar.close()
+    return origin_array
+
+
+def polar_transform(
```
bobleesj, line 281: i believe this is the main function

for structure the order of util code/functions, we want to order functions in a hierarchical manner - start with an overview and details. 

right now, I find this main public API function in line 460, and I have to climb up to see that all these private compoennts are consisted of

ref: https://github.com/electronmicroscopy/quantem/pull/208/changes#diff-6b1831e12ac34971924eaa5de100c41a69999a64dea70f15504ee13303b5bd68

bobleesj (commented):
@ehrhardtkm This is much improved - thanks for adding docstrings and conventions so that it's easier for me read and follow. I spent 20-30 mins reviewing this code. I will run with real data in the next 2-3 days or so. Please feel free to address my comments in the meantime.

src/quantem/diffraction/polar.py
bobleesj: okay! most you did :)

bobleesj:
@ehrhardtkm 


Architetural discussion - regarding

```
ds = em.core.io.read_4dstem("path/to/data.h5", file_type="arina")

# Separate steps:
origins = ds.auto_origin_id()
polar = ds.polar_transform(origin_array=origins)
pdf = PairDistributionFunction.from_data(polar)
```

I have a question regarding the scope of `dataset4dstem` -  I'm wondering whether it should have `auto_origin_id` and `polar_transform` included in this dataset class

To me, these origin finding and polar transforms are the role of `PairDistributionFunction` instead of the role of `dataset4dstem` as these are currently used _once_ in the entire codebase atm.

This way, we try to minimize any side effects done on these datasets classes (they do just one thing very well - manipulate/load/save **raw** data. Any derived data I feel like it's not the role of `dataset` classes because then the role becomes diluted.

bobleesj:
[PDFanalysis_simulatedTa_250504_api.ipynb](https://github.com/user-attachments/files/27379340/PDFanalysis_simulatedTa_250504_api.ipynb)

Updated notebook API. Did testing on my laptop and it runs end-to-end. I have gold data so it can be really fun to play along with your widget soon interactively.

src/quantem/diffraction/polar.py
ehrhardtkm: Yes, this is standard. I added a line to let new users know where they can learn more about the notation in the docstring

src/quantem/diffraction/polar.py
ehrhardtkm: Done

src/quantem/diffraction/polar_transform.py
ehrhardtkm: Reordered functions as suggested. Main API functions are at the top now and helper functions are at the bottom.

src/quantem/diffraction/polar_transform.py
ehrhardtkm: Removed `_precompute_polar_coords` entirely and replaced it a single call to `_build_polar_sampling_offsets` that can be reused by translating offsets to the absolute origin px location for each origin. This is also faster to run!

src/quantem/diffraction/polar.py
ehrhardtkm: Added this info to the docstring, along with usage example section

src/quantem/diffraction/polar.py
ehrhardtkm: Added to docstring

src/quantem/diffraction/polar.py
ehrhardtkm: Tried to make sure it was numpy style, please let me know if I'm missing something

src/quantem/diffraction/polar.py
ehrhardtkm: Done

src/quantem/tomography/utils.py
ehrhardtkm: I agree that the PDF module shouldn't use tomo utils, but the use case for this function isn't for imaging either (it's for the baspadd filtering to clean up noise in the structure factor). I'm okay with moving this there for now, but I'll also need to copy over the `gaussian_kernel_1d` function too

src/quantem/diffraction/polar.py
ehrhardtkm: This should now just accept `dataset4dstem` or `dataset2d`. Currently, `dataset4dstem` only takes numpy arrays, so the broader goal of making the code robust to different data types is probably an issue outside of the scope of this PR and maybe something to have a larger conversation about with others.

ehrhardtkm:
In response to the question about architecture: with the last refactor, I moved `auto_origin_id` and `polar_transform` out of `Dataset4dstem` precisely for this reason. They're now top-level functions in `polar_transform.py`. I could maybe put them in `Polar4dstem` if you think that makes more sense. The new usage looks like:
```
origins = auto_origin_id(ds)
polar = polar_tranform(ds, origins)
```

ehrhardtkm:
Here's an updated notebook: [PDFanalysis_simulatedTa.ipynb](https://github.com/user-attachments/files/27420397/PDFanalysis_simulatedTa.ipynb)

src/quantem/tomography/utils.py
ehrhardtkm: @bobleesj and I discussed in person that this can stay in tomo utils fo now to avoid duplicating code in the repo, but we should have a bigger conversation about moving some of these functions to a general utils file soon.

ehrhardtkm:
@bobleesj Turns out that my branch had diverged from a tomo refactor, so the gaussian filter functions that I need were actually moved to `core/util/utils.py`. I added my `gaussian_filter_1d` function to that file and changed imports accordingly. Everything in the tomo module should be identical to upstream.dev.

This also fixed the CI issue, so we got two birds with one scone!

bobleesj (commented):
@ehrhardtkm thank you Karen for all the updates. merging it to `upstream/nanobeam`. 

cc' @cophus

PR #180, opened 2026-02-22
Updating strain_autocorrelation to fix dft upsampling and add multi-peak cepstral strain mapping

mitis1:

Updates the strain_autocorrelation function to fix dft_upsampling for accurate subpixel peak fitting. Adds the option of multi-peak refinement of u and v vectors of cepstral transform for more accurate strain mapping. The attached test notebook is an example test of just the multi-peak cepstral, not the full strain autocorrelation function. It uses an example dataset and runs a multi-peak cepstral fit of strain for it. 
[multi-peak-cepstral-test.ipynb](https://github.com/user-attachments/files/25465824/multi-peak-cepstral-test.ipynb)

Test multi-peak fitting code with different parameters. 

Check that the multi-peak fitting works on a test dataset and doesn't break any of the other functions in strain-autocorrelation.py. Verify that the dft-upsampling works and improves peak fitting of u and v vectors.


bobleesj:
@mitis1 I believe this PR is closed since it's coming from `mitis1:nanobeam`?

PR #184, opened 2026-03-03
Overhaul of vector

cophus:
# What this PR does

This PR completely rewrites the `Vector` class.

The new implementation is substantially simpler. Public proxy types such as `FieldView` and `CellView` are removed, and all indexing/selection operations now return `Vector`. The explicit NumPy escape hatches are:
- `.flatten()` for concatenated rowwise data
- `.array` for selections containing exactly one cell

The main syntax change is field selection. Instead of:
`v["kx"]`
use:
`v.select_fields("kx")`

This makes fixed-grid indexing with `[]` distinct from field selection.

`select_fields(...)` returns a write-through `Vector` view, so field-selected updates still modify the parent `Vector`.

This PR also adds:
- rowwise update helpers such as `set_flattened(...)`
- `append_rows(...)` and `compact()`
- expanded arithmetic support, including `**`, `%`, `//`, unary `-`, `abs(...)`
- elementwise NumPy ufunc support such as `np.sin(vector)` and `np.maximum(vector, 0)`
- Minor fix for the 2D histogram calculation function.

File I/O is also much cleaner. The internal storage no longer serializes nested per-cell containers, which previously created large Zarr metadata overhead. The new layout stores one numeric row buffer plus per-cell offsets/lengths. In practice this makes `Vector` saves substantially smaller and reduces read/write overhead.

# What PR reviewer(s) should do

- Run the `Vector` test suite and confirm downstream tests still behave as expected.
- Run the attached notebook examples.
- Call out any missing functionality or awkward syntax before this API is finalized.

# Notebook demo for syntax:
[vector_new_01.ipynb](https://github.com/user-attachments/files/25698793/vector_new_01.ipynb)

### DIFF
--- a/src/quantem/core/datastructures/vector.py
+++ b/src/quantem/core/datastructures/vector.py
@@ -1,8 +1,7 @@
 from __future__ import annotations
 
 import copy
-from dataclasses import dataclass
-from typing import Any, Iterable, Sequence, overload
+from typing import Any, Sequence, overload
 
 import numpy as np
 from numpy.typing import NDArray
@@ -16,23 +15,13 @@ from quantem.core.utils.validators import (
 )
 
 
-@dataclass
-class _VectorState:
-    shape: tuple[int, ...]
-    data: NDArray[np.object_]
-    fields: list[str]
-    units: list[str]
-    name: str
-    metadata: dict[str, Any]
-
-
 class Vector(AutoSerialize):
     """Ragged cell data on a fixed grid.
 
-    A ``Vector`` has fixed grid dimensions (``shape``). Each fixed-grid cell stores a
-    NumPy array with shape ``(n_rows, num_fields)``. Selections always return new
-    ``Vector`` views over the same backing store, while raw NumPy extraction is explicit
-    via ``.array`` and ``.flatten()``.
+    Storage is compact and AutoSerialize-friendly:
+    - one 2D numeric row buffer for all ragged rows
+    - one start array and one length array for cell boundaries
+    - optional selection state for views
     """
 
     def __init__(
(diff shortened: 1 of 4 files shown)


src/quantem/core/datastructures/vector.py
```diff
@@ -1,1026 +1,1217 @@
...
+    The internal representation is compact:
+    - ``_state["data"]`` stores all ragged rows in one numeric 2D array
+    - ``_state["cell_starts"]`` stores the start offset for each cell
+    - ``_state["cell_lengths"]`` stores the row count for each cell
+
+    A ``Vector`` selection is a write-through view over shared storage. Views
+    track only the selected fixed-grid shape, selected cell indices, and selected
+    field names.
+
+    Examples
```
bobleesj, line 63: beautiful...

tests/datastructures/test_vector.py
```diff
@@ -1,433 +1,407 @@
...
+    v = Vector.from_shape(shape=(3, 2), fields=["intensity", "kx", "ky"])
+    for i in range(3):
+        for j in range(2):
+            base = float(i * 10 + j)
+            v[i, j] = np.array([[base, base + 100.0, base + 200.0]])
+    return v
+
+
+class TestVector:
+    def test_initialization_and_len(self):
```
bobleesj, line 34: just like the examples section in the API, this testsfurther nicely shows *how* the vector class works

arthurmccray:
Okay this looks good to me. I reckon there will be some changes that need to be made as this gets used in features, but it does everything I want for now at least. 

It's not the easiest to understand all of the code, but unless we start subclassing it, then I don't think that should ever be a problem. The tests we have are fairly robust I think.

I also created a tutorial at https://github.com/electronmicroscopy/quantem-tutorials/pull/5

arthurmccray (approved):

PR #190, opened 2026-03-10
ptycho: enhance `show_scan_positions` for probe overlap/size estimation and metadata

bobleesj:
### What this PR does

No physics is changed for probe size estimation. 

User wants to get an approximation for probe overlap, size, etc. The existing implementation works but with a minor fix as mentioned in the in-line comment:

Here is the plot and output:

```
ptycho_pix.show_scan_positions(num_probes=20)
```

Screenshot 2026-03-09 at 6 25 08 PM

Yes, probes are tiny.

API renders:

Screenshot 2026-03-09 at 6 22 14 PM

This was brought up while going through the ptycho notebook w/ @arthurmccray. I will probably the only user for now but since no physics is modified, it can be merged but if you more ideas, please share.

### DIFF
--- a/src/quantem/diffractive_imaging/ptychography_visualizations.py
+++ b/src/quantem/diffractive_imaging/ptychography_visualizations.py
@@ -1,3 +1,4 @@
+
 import warnings
 from typing import Any, Literal
 
@@ -886,7 +887,7 @@ class PtychographyVisualizations(PtychographyBase):
         scan_sampling = self.dset.scan_sampling
         fov = scan_sampling * (np.array(scan_gpts) - 1)
         print(f"Scan grid: {scan_gpts[0]} x {scan_gpts[1]} positions")
-        print(f"FOV: {fov[0]:.2f} x {fov[1]:.2f} Å")
+        print(f"Object FOV: {fov[0]:.2f} x {fov[1]:.2f} Å")
 
         if conv_angle is not None and energy is not None:
             from quantem.core.utils.utils import electron_wavelength_angstrom
--- a/src/quantem/diffractive_imaging/ptychography_visualizations.py
+++ b/src/quantem/diffractive_imaging/ptychography_visualizations.py
@@ -829,50 +829,89 @@ class PtychographyVisualizations(PtychographyBase):
             **kwargs,
         )
 
-    def show_scan_positions(self, plot_radii: bool = True):
-        """
-        Show the scan positions and the probe radius.
+    def show_scan_positions(
+        self,
+        plot_radii: bool = True,
+        num_probes: int | None = None,
+        axsize: tuple[int, int] = (10, 10),
+        edgecolors: str = "red",
+        linewidths: float = 0.5,
+        **kwargs,
+    ):
+        r"""Show scan positions, probe radii, and overlap metadata
+
+        Visualize probe coverage to verify sufficient overlap for
+        ptychographic reconstruction. The probe radius is estimated as
+        :math:`r = 0.61 \lambda / \alpha + |\Delta f| \cdot \alpha`
+        where :math:`\lambda` is the electron wavelength, :math:`\alpha` is
+        the convergence semi-angle, and :math:`\Delta f` is the defocus.
+        Overlap is :math:`(d - s)/d = 1 - s/d` where :math:`d = 2r` is
+        the probe diameter and :math:`s` is the scan step size.
+
         Parameters
         ----------
-        plot_radii: bool, optional
-            Whether to plot the probe radius, by default True
+        plot_radii : bool, optional
+            Whether to plot the probe radius. Default is True
+        num_probes : int | None, optional
+            Number of probe positions to display. Default is 3 rows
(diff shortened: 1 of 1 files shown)


src/quantem/diffractive_imaging/ptychography_visualizations.py
```diff
@@ -829,50 +829,94 @@ def _show_object_and_probe_iters(
...
-            diffraction_limit_angstrom = 0.61 * wavelength / conv_angle_rad
-            defocus_blur_angstrom = abs(defocus) * conv_angle_rad
-            probe_radius_angstrom = diffraction_limit_angstrom + defocus_blur_angstrom
-            probe_radius_px = probe_radius_angstrom / self.sampling[0]
-
-        _fig, ax = show_2d(self._get_probe_overlap(), title="probe overlap")
+            diffraction_limit_A = 0.61 * wavelength / conv_angle_rad
+            defocus_blur_A = abs(defocus) * conv_angle_rad
+            probe_radius_A = diffraction_limit_A + defocus_blur_A
+            probe_radius_px = probe_radius_A / self.sampling[0]
```
bobleesj, line 900: I noticed that the ptycho sampling is different from the original sampling.

For printing, I used the original sampling

bobleesj (commented):
@arthurmccray ready for review

src/quantem/diffractive_imaging/ptychography_visualizations.py
```diff
@@ -829,50 +829,94 @@ def _show_object_and_probe_iters(
...
         probe_radius_px = None
-
         conv_angle = probe_params.get("semiangle_cutoff")
         defocus = probe_params.get("defocus", 0)
         energy = probe_params.get("energy")
+        scan_gpts = self.dset.gpts
+        scan_sampling = self.dset.scan_sampling
+        fov = scan_sampling * (np.array(scan_gpts) - 1)
+        print(f"Scan grid: {scan_gpts[0]} x {scan_gpts[1]} positions")
+        print(f"FOV: {fov[0]:.2f} x {fov[1]:.2f} Å")
```
bobleesj: Thought might be useful to show FOV - while lenghh bar helps, FOV gives more like a "global" yard stick.

src/quantem/diffractive_imaging/ptychography_visualizations.py
```diff
@@ -829,50 +829,94 @@ def _show_object_and_probe_iters(
...
             conv_angle_rad = conv_angle * 1e-3
-
             # For defocused probe: radius ≈ |defocus| * convergence_angle + diffraction_limit
-            diffraction_limit_angstrom = 0.61 * wavelength / conv_angle_rad
-            defocus_blur_angstrom = abs(defocus) * conv_angle_rad
-            probe_radius_angstrom = diffraction_limit_angstrom + defocus_blur_angstrom
-            probe_radius_px = probe_radius_angstrom / self.sampling[0]
-
-        _fig, ax = show_2d(self._get_probe_overlap(), title="probe overlap")
+            diffraction_limit_A = 0.61 * wavelength / conv_angle_rad
```
bobleesj, line 897: felt `angstrom` a bit too long since I was repeating alot while I dind't want to lose the unit. So I used `A` but `ang` would have worked too. Not sure what our convention is.

bobleesj (commented):
A few more comments on unit code convention and metadata

src/quantem/diffractive_imaging/ptychography_visualizations.py
arthurmccray: This is the FOV of the object rather than the FOV of the image that's currently being shown with the probe positions, do you think that's confusing? Maybe relabel to `f"Object FOV: ...`

src/quantem/diffractive_imaging/ptychography_visualizations.py
arthurmccray: `ptycho.sampling` is the realspace sampling of the reconstructed object, i.e. it's dependent on the maximum scattering angle of the diffraction patterns. We generate this image with the scan positions using the `_get_probe_overlap` method, which uses the same patches backend as is in the forward model, so that is the correct sampling to use. Just wanted to clarify :)

arthurmccray (approved):
Looks good! I would make the super minor change of specifying the object FOV in the printed string, but otherwise ready to merge.

src/quantem/diffractive_imaging/ptychography_visualizations.py
bobleesj: yes, thank you - done.

src/quantem/diffractive_imaging/ptychography_visualizations.py
bobleesj: thanks for this clafirication! I will take a look!

bobleesj (commented):
@arthurmccray Thanks Arthur for your feedback on adding "FOV" text. This has been fixed. Ready for review and merge.

PR #191, opened 2026-03-10
enhance `bin` and `crop` API docstring with examples 

bobleesj:
### What this PR does

`crop` and `bin` are commonly used during the first few cells of 4D-STEM data processing.

While its usage can be learned from the notebooks, this PR adds examples to the documentation so users can use it w/o searching through notebooks (experienced by @bobleesj since RAM is precious)

### What should reviewer(s) do

No function changed. Just enhancing the doc. Please take a quick look at the in-line.

### DIFF
--- a/src/quantem/core/datastructures/dataset.py
+++ b/src/quantem/core/datastructures/dataset.py
@@ -452,21 +452,56 @@ class Dataset(AutoSerialize):
         axes: tuple | None = None,
         modify_in_place: bool = False,
     ) -> Self | None:
-        """
-        Crops Dataset
+        """Select a sub-region of the dataset along specified axes
+
+        Unlike regular slicing (``dset[64:192, ...]``), which creates a
+        view that keeps the full original array in memory,
+        ``modify_in_place=True`` replaces the array and frees the original.
+        This is important for large 4D-STEM datasets where memory is
+        limited.
+
+        Each ``crop_widths`` entry is a ``(start, stop)`` pair defining
+        which elements to keep. A ``stop`` of ``0`` keeps everything from
+        ``start`` to the end.
 
         Parameters
         ----------
-        crop_widths:tuple
-            Min and max for cropping each axis specified as a tuple
-        axes:
-            Axes over which to crop. If None specified, all are cropped.
-        modify_in_place: bool
-            If True, modifies dataset
+        crop_widths : tuple[tuple[int, int], ...]
+            ``(start, stop)`` indices for each axis specified in ``axes``.
+        axes : tuple | None
+            Axes to crop. If None, all axes are cropped.
+        modify_in_place : bool
+            If True, modifies this dataset in-place and frees the original
+            array. If False, returns a new dataset.
 
         Returns
+        -------
+        Dataset | None
+            Cropped dataset if ``modify_in_place`` is False, otherwise None.
+
+        Examples
         --------
-        Dataset (cropped) only if modify_in_place is False
+        Crop real-space to a 128x128 region:
+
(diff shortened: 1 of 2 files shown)


src/quantem/core/datastructures/dataset.py
```diff
@@ -452,21 +452,56 @@ def crop(
...
         Parameters
         ----------
-        crop_widths:tuple
-            Min and max for cropping each axis specified as a tuple
-        axes:
-            Axes over which to crop. If None specified, all are cropped.
-        modify_in_place: bool
-            If True, modifies dataset
+        crop_widths : tuple[tuple[int, int], ...]
+            ``(start, stop)`` indices for each axis specified in ``axes``.
```
bobleesj, line 464: used `start` `stop`, matching Python API convention - `range(start, stop, step)`

src/quantem/core/datastructures/dataset.py
```diff
@@ -452,21 +452,56 @@ def crop(
         axes: tuple | None = None,
         modify_in_place: bool = False,
     ) -> Self | None:
-        """
-        Crops Dataset
+        """Select a sub-region of the dataset along specified axes
+
+        Unlike regular slicing (``dset[64:192, ...]``), which creates a
```
bobleesj: Addressing the "why" we use `crop`

src/quantem/core/datastructures/dataset.py
```diff
@@ -526,27 +561,51 @@ def bin(
...
+            Binned dataset if ``modify_in_place`` is False, otherwise None.
 
         Notes
         -----
         - Any remainder (shape % factor) is dropped on each binned axis.
         - Sampling is multiplied by the factor on each binned axis.
         - Origin is shifted to the center of the first block:
             origin_new = origin_old + 0.5 * (factor - 1) * sampling_old
+
+        Examples
```
bobleesj, line 591: 2 examples for binnning - covers most of the cases.

src/quantem/core/datastructures/dataset.py
```diff
@@ -452,21 +452,56 @@ def crop(
...
+        modify_in_place : bool
+            If True, modifies this dataset in-place and frees the original
+            array. If False, returns a new dataset.
 
         Returns
+        -------
+        Dataset | None
+            Cropped dataset if ``modify_in_place`` is False, otherwise None.
+
+        Examples
```
bobleesj, line 476: 3 examples

1. crop real-space and view
2. crop k-space and view
3. crop k-space and modifiy in place

bobleesj (commented):
@arthurmccray ready for review - I think I asked you before what was the best way to crop 4dstem dataset so that I can fit stuff in VRAM. I am sure there will be another @bobleesj down the road.

bobleesj:
I was also looking at whether `modifiy_in_place=true` should free the initial memory. not sure whether it's happening in the current design. 

But I think being able to free the original array makes this `crop` function more useful than using native slicing methods (which returns a view)??

src/quantem/core/datastructures/dataset.py
bobleesj: (outdated -> noticed we don't actually release the memory, but needs further investigaiton)

arthurmccray (approved):
lgtm!

regarding `modify_in_place=True`, memory profiling can be a little annoying in python, but I'm pretty sure that this should work as expected and free the memory. And it's nice having both this and the slicing methods so we have the option for views as well :)

bobleesj:
@arthurmccray Thanks Arthur for your quick review. If it looks ok! please merge! I will just double check whether it frees the memory!

PR #196, opened 2026-03-19
initializing spectroscopy 

smribet:
initializing spectroscopy. more details coming soon.


src/quantem/core/datastructures/dataset1d.py
```diff
@@ -0,0 +1,216 @@
...
+            sampling=sampling,
+            units=units,
+            signal_units=signal_units,
+        )
+
+    def show(
+        self,
+        title: str | None = None,
+        returnfig: bool = False,
+        **kwargs,
```
cophus: **kwargs presumably should be passed through to plot?

src/quantem/core/io/file_readers.py
```diff
@@ -9,6 +9,13 @@
 from quantem.core.datastructures import Dataset2d as Dataset2d
 from quantem.core.datastructures import Dataset3d as Dataset3d
 from quantem.core.datastructures import Dataset4dstem as Dataset4dstem
+from quantem.spectroscopy import (
+    Dataset3deds as Dataset3deds,
```
cophus: Are we agreed on naming it "eds?"  I prefer the more explicit name "xeds" i.e. x-ray energy dispersive spectrometry.

src/quantem/core/io/file_readers.py
```diff
@@ -147,6 +154,96 @@ def read_4dstem(
...
+    # If specific index provided, use it
+    if dataset_index is not None:
+        imported_data = data_list[dataset_index]
+        if imported_data["data"].ndim != 3:
+            raise ValueError(
+                f"Dataset at index {dataset_index} has {imported_data['data'].ndim} dimensions, "
+                f"expected 4D. Shape: {imported_data['data'].shape}"
+            )
+    else:
+        # Automatically find first 3D dataset
```
cophus, line 195: We could add a function to print out the dataset indices, shapes, and other metadata line names if available.

src/quantem/spectroscopy/dataset3deds.py
```diff
@@ -0,0 +1,3643 @@
...
+    PolynomialBackground,
+    abundance_smoothness_l2,
+    build_element_basis,
+    eds_data_loss,
+    inverse_softplus,
+    polynomial_energy_basis,
+)
+
+
+class Dataset3deds(Dataset3dspectroscopy):
```
cophus: As mentioned above, maybe should be:
Dataset3dxeds

src/quantem/spectroscopy/dataset3deds.py
```diff
@@ -0,0 +1,3643 @@
...
+            **getattr(self, "_spectrum_images", {}),
+            **dict(zip(labels, maps)),
+        }
+
+        images, titles = self.show_spectrum_images(x_ray_lines=elements, return_maps=True)
+
+        if return_maps:
+            return images, titles
+
+    def Integrate(self, spec, width=0.15, return_maps=False, show=True, **kwargs):
```
cophus: I don't think we should support "Integrate" and "integrate." Generally methods should all be lowercase.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+                type(self)._resolve_element_from_label(lbl, ordered_elements) for lbl in sel_labels
+            } - {None}
+            if len(matched_elements) != 1:
+                raise ValueError(
+                    f"Selector {selector!r} matched multiple elements: {sorted(matched_elements)}"
+                )
+            element = next(iter(matched_elements))
+
+            grouped_map = np.sum([line_map[lbl] for lbl in sel_labels], axis=0)
+            intensity = float(grouped_map.sum())
```
cophus, line 688: these "float()" wrappers are not necessary.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        return requested or saved
+
+    @staticmethod
+    def _estimate_snr_thresholds(snr_values, floor=None, snr_threshold=None):
+        """Auto-estimate SNR floors/thresholds from the peak SNR distribution."""
+        snr_values = np.asarray(snr_values, dtype=float)
+        snr_values = snr_values[np.isfinite(snr_values)]
+
+        if floor is None:
+            if snr_values.size:
```
cophus: This seems quite arbitrary - perhaps user could override these values as an arg?

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        fig, ax = show_2d(
+            images,
+            title=titles,
+            cmap=kwargs.pop("cmap", "magma"),
+            scalebar={"sampling": self.sampling[1], "units": self.units[1]},
+            returnfig=True,
+            **kwargs,
+        )
+
+        if return_fig and return_maps:
```
cophus: We should return Dataset2d - this way metadata such as pixel size and which element line(s) the map was calculated from can be captured.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+            result.update(
+                {
+                    "selector_maps": selector_maps,
+                    "intensity_maps": intensity_maps,
+                    "weighted_intensity_maps": weighted_intensity_maps,
+                    "atomic_percent_maps": atomic_percent_maps,
+                    "weight_percent_maps": weight_percent_maps,
+                }
+            )
+        return result
```
cophus: I think returning dicts of numpy arrays is somewhat not in the spirit of quantEM. Using Dataset3d / Dataset2d allows us to keep sampling values (pixel size) and metadata. Also gives us the ability use the .save() method for the outputs.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        def collapse_shoulder_peaks(indices, heights, prominences, widths):
+            if len(indices) <= 1:
+                return (
+                    np.asarray(indices, dtype=int),
+                    np.asarray(heights, dtype=float),
+                    np.asarray(prominences, dtype=float),
+                    np.asarray(widths, dtype=float),
+                )
+
+            energy_gap_limit = max(6.0 * float(self.sampling[2]), 0.14)
```
cophus: This should be an arg for the user to override if needed.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        snr_values = np.asarray([height / background_std for height in peak_heights], dtype=float)
+        floor, snr_threshold = type(self)._estimate_snr_thresholds(
+            snr_values,
+            floor,
+            snr_threshold,
+        )
+
+        # Prominence filter in SNR units: suppress shoulder/noise artifacts that
+        # may have acceptable height but do not form a distinct peak.
+        prominence_snr = np.asarray(
```
cophus: This might be brittle if the spectrum has been denoising. Background std could be very low.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+
+        gauss_components = []
+        for pk_idx, _, _, snr, _, _ in display_peaks_with_prom:
+            if float(snr) >= gaussian_validation_gate:
+                continue
+            fit = _fit_local_gaussian(int(pk_idx))
+            if fit is None:
+                continue
+            # Keep only physically plausible and sufficiently Gaussian components.
+            if fit["r2"] < 0.58:
```
cophus: Not clear where these arbitrary thresholds came from.  It's ok to use some "reasonable" values but this should be documented in comments and / or method comment block.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+                if eliminate:
+                    break
+            if eliminate:
+                demoted_elements.add(str(element))
+
+        element_confidence = {}
+        # --- Intensity ratio check and multi-peak pattern boost ---
+        for element, stats in element_stats.items():
+            valid_shells = {shell for shell in stats["shells"] if shell in {"K", "L", "M"}}
+            shell_bonus = float(np.sqrt(max(1, len(valid_shells))))
```
cophus: This also seems very arbitrary!  Should be explained.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        energy_axis_np = self.energy_axis.copy()
+        energy_axis = torch.tensor(energy_axis_np, dtype=torch.float32, device=device)
+        spectra = torch.tensor(self.array, dtype=torch.float32, device=device)
+
+        if energy_range is not None:
+            ind = (energy_axis >= energy_range[0]) & (energy_axis <= energy_range[1])
+            energy_axis = energy_axis[ind]
+            spectra = spectra[:, :, ind]
+        else:
+            energy_range = [float(energy_axis.min().item()), float(energy_axis.max().item())]
```
cophus: another example of float( ) not needed

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        )
+        ax[1].set_xlim(energy_range[0], energy_range[1])
+        ax[1].legend()
+        ax[1].set_title("fit spectrum")
+        ax[1].set_xlabel("Energy (keV)")
+        ax[1].set_ylabel("Counts")
+        plt.tight_layout()
+        plt.show()
+
+        return {
```
cophus: Should we define a quantEM dict class which can be AutoSerialize'd?

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+            "element_names": model.peak_model.element_names,
+            "edge_concentrations": shell_concs,
+            "edge_names": shell_names,
+            "edge_element_indices": shell_element_indices,
+            "peak_widths": final_fwhm,
+            "energy_axis": energy_axis_plot,
+            "fit_range": energy_range,
+        }
+
+    def fit_spectrum_pytorch(
```
cophus: This function should probably be moved over to the shared model fitting architecture defined by Arthur. This can be done later however.

src/quantem/spectroscopy/dataset3deds.py
cophus: This class is too long.  I suggest some of the processing code should be moved into another py file, perhaps xeds_models.py or xeds_fitting.py. vis functions could also be moved into another .py file.

Modeling components should (eventually) we re-tooled to use the same core functions created here:
https://github.com/electronmicroscopy/quantem/tree/fitting_models_clean

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+from scipy.interpolate import interp1d
+from scipy.ndimage import median_filter
+from scipy.optimize import curve_fit
+from scipy.stats import norm
+
+from quantem.core.visualization import show_2d
+from quantem.spectroscopy.dataset3dspectroscopy import Dataset3dspectroscopy
+
+
+class Dataset3deels(Dataset3dspectroscopy):
```
cophus, line 34: Same comment as dataset3deds: Fitting functions and vis functions could be split off into other .py files for easier navigation and readability.

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+    """An EELS dataset class that inherits from Dataset3dspectroscopy.
+
+    This class represents a scanning transmission electron microscopy (STEM) dataset,
+    where the data consists of a 3D array with dimensions (scan_y, scan_x, energy).
```
cophus: Here the data is ordered (row,column,energy) which is what the devs voted for. However, the docstring says (scan_y, scan_x, energy), but measure_zlp_offset below unpacks shape[0] as n_x, so the two disagree on what the first axis is. Suggest committing to row/col naming to remove the ambiguity:

`where the data consists of a 3D array with dimensions (scan_row, scan_col, energy).`

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        -------
+        ndarray
+            Background-subtracted spectrum
+        """
+
+        from scipy.ndimage import gaussian_filter
+        from scipy.stats import norm
+
+        # Smooth for better fitting
+        spec_smooth = gaussian_filter(spectrum, sigma=1.0)
```
cophus: Another arbitrary smoothing value - needs to be a user-selectable arg.

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        # Iteratively fit Gaussian to low-intensity values (the continuum)
+        # Remove outliers (edge peaks) iteratively
+        num_iterations = 10
+        cutoff = 3  # +/- 3 sigma
+
+        for _ in range(num_iterations):
+            mu, std = norm.fit(pixel_vals)
+            if std == 0:
+                break
+            # Keep only values within +/- 3 sigma (removes edge contributions)
```
cophus: Isn't this +/- 1 std?

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        if target_edge < energy_axis[0] or target_edge > energy_axis[-1]:
+            raise ValueError("Target edge is outside of energy range.")
+        elif ((target_edge - 5) - target_edge * (window_size / 100)) < energy_axis[0]:
+            raise ValueError(
+                "Insufficient pre-edge background fitting region for this target edge and window size within given energy range."
+            )
+
+        # Fit power law function to spectrum within window region of the energy exis
+
+        window_minE = (target_edge - 5) - target_edge * (window_size / 100)
```
cophus: window_min_E would be less confusing, this looks like a typo

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+            ymax=np.max(spectrum),
+            label="window limits",
+            color="k",
+            linestyle="dashed",
+        )
+        ax.legend()
+
+        return background_fit
+
+    def smooth_eels_rollingaverage(self, roi=None, energy_range=None, mask=None, kernel_size=10):
```
cophus: another strange choice for underscores, could be:
smooth_eels_rolling_average

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+                "Check that zlp_shifts_array is in energy units, not channel indices."
+            )
+
+        new_origin = float(new_Eaxis[0])
+
+        # Calculate mean spectra before and after correction for plotting
+        mean_spectrum_raw = self.array.mean(axis=(0, 1))
+        mean_spectrum_corrected = aligned_data_3d.mean(axis=(0, 1))
+
+        fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 4))
```
cophus, line 793: Some of the plotting infrastructure could be shared between XEDS and EELS. Perhaps a spectroscopy_vis.py function?

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        y_coords, x_coords = np.meshgrid(np.arange(n_y), np.arange(n_x), indexing="ij")
+        y_flat = y_coords.ravel()
+        x_flat = x_coords.ravel()
+        z_flat = zlp_map.ravel()
+
+        A = np.column_stack([y_flat, x_flat, np.ones(len(y_flat))])
+        coeffs, _, _, _ = np.linalg.lstsq(A, z_flat, rcond=None)
+        a, b, c = coeffs
+
+        zlp_plane = a * y_coords + b * x_coords + c
```
cophus: Are linear plane fits enough?  Or will we need higher order polynormials / bezier polynomials for some datasets?

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+            return Dataset3deels.from_array(
+                array=aligned_data_3d,
+                name=self.name,
+                sampling=self.sampling,
+                origin=new_origin,
+                units=self.units,
+            )
+        return aligned_data_3d
+
+    def calibrate_zero_loss_peak(self, center_guess=None, search_window=10):
```
cophus: Suggest adding an arg
`in_place = False `

This way we can prevent needing 2x RAM in some cases.

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+
+        print(f"QuantEM: Calculating thickness for {dataset.name}...")
+
+        # 1. Vectorized Integration
+        I_zlp = np.sum(data[:, :, z_start : z_end + 1], axis=2)
+        I_total = np.sum(data[:, :, t_start : t_end + 1], axis=2)
+
+        # 2. Log-Ratio Calculation (with epsilon to avoid log(0))
+        epsilon = 1e-10
+        t_over_lambda = np.log((I_total + epsilon) / (I_zlp + epsilon))
```
cophus: could use np.log1p() as an alternative to episilon.

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+import torch
+from matplotlib.patches import Rectangle
+from numpy.typing import NDArray
+
+from quantem.core.datastructures.dataset3d import Dataset3d
+from quantem.core.visualization import show_2d
+from quantem.spectroscopy.utils import load_eels_edges_database, load_xray_lines_database
+
+
+class _ModelElementsDict(dict):
```
cophus, line 34: Is it worth putting this upstram in quantem core?

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        dict
+            Dictionary containing:
+            - 'pca': PCA result attributes
+            - 'components': principal component spectra (n_components x n_energy)
+            - 'loadings': spatial loadings (n_components x scan_y x scan_x)
+            - 'explained_variance_ratio': explained variance for each component
+            - 'reconstructed': reconstructed dataset (dataset3dspectroscopy) using n_components
+        """
+
+        from quantem.spectroscopy import Dataset3deds, Dataset3deels
```
cophus: Minor / API-consistency thought: rather than importing the concrete classes, could we route array construction through a single dispatched entry point, the way read_3d_spectroscopy(..., data_type="EELS") already does for files? Something like:

```
# quantem/spectroscopy/__init__.py
def from_array(array, data_type, **kwargs):
    cls = {"eels": Dataset3deels, "eds": Dataset3deds}[data_type.lower()]
    return cls.from_array(array, **kwargs)
```

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        ----------
+        normalize : bool, optional
+            If ``True``, scale the mean spectrum to the range [0, 1]. If
+            ``False``, return the mean spectrum in original intensity units.
+        """
+        y, x, dy, dx = self._resolve_roi(roi=roi, roi_cal=roi_cal)
+
+        # SPECTRUM CALCULATION --------------------------------------------------------------
+
+        E = np.asarray(self.energy_axis, dtype=float)
```
cophus: E = self.energy_axis
is fine

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        fig.tight_layout()
+
+        if show:
+            plt.show()
+
+        return fig, (ax_map, ax_spec), energy_map
+
+    # BACKGROND SUBTRACTION
+
+    def subtract_background(
```
cophus, line 752: I think we need more complex background models besides just global and local. Specifically we want to be able to subtract backgrounds / use fitting models where the peak positions are fixed and only intensity changes. Or for backgrounds, power laws defined by:
constant * energy ** power
where power = fixed, but constant can vary.

src/quantem/spectroscopy/eels_edges.csv
```diff
@@ -0,0 +1,1047 @@
+atomic_number,symbol,element,edge_label,edge_energy_eV
```
cophus, line 3: Where are these from? Should we have a citation?

src/quantem/spectroscopy/x_ray_lines.csv
```diff
@@ -0,0 +1,702 @@
+energy_eV,atomic_number,element,line,relative_intensity,weight
```
cophus, line 3: Where are these from? Do we need a citation?

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+
+        # Define Gaussian constraint to fit ZLP to
+        def _gaussian_fit(x, A, mu, sigma):
+            return A * np.exp(-0.5 * ((x - mu) / sigma) ** 2)
+
+        def _plane_fit_2d(M, a, b, c):
+            x, y = M
+            return (a * x) + (b * y) + c
+
+        n_x, n_y, _n_energy = self.array.shape
```
cophus: Here shape[0] is n_x, shape[1] is n_y, but calibrate_zero_loss_peak (line 470) uses the opposite (shape[0] is n_y). Both are internally correct, but the clash is confusing. Could we standardize to scan_row, scan_col here and propagate to the loop vars (:269-270), the indexing (:273, :275), and the reshape at :329? Same convention as the docstring above.

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+            Number of channels to search on either side of center_guess.
+            Only used when center_guess is not None. Default is 10.
+
+        Returns
+        -------
+        Dataset3deels
+            New dataset with corrected energy calibration.
+        """
+
+        n_y, n_x, n_energy = self.array.shape
```
cophus: Mirror of the comment on measure_zlp_offset: this method names shape[0] as n_y while that one names it n_x. Picking scan_row, scan_col consistently across both (and the loop/index/meshgrid lines :490-506) would make them match. No behavior change — just readability.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+            {(lbl, round(float(e), 12), round(float(w), 12)) for lbl, e, w in rows},
+            key=lambda t: (t[1], -t[2], t[0]),
+        )
+        return (
+            np.asarray([e for _, e, _ in unique], dtype=float),
+            np.asarray([w for _, _, w in unique], dtype=float),
+            [lbl for lbl, _, _ in unique],
+        )
+
+    def generage_spectrum_images(self, elements=None, width=0.15, return_maps=False):
```
cophus: Typo in this public method name (generage → generate). Worth fixing now before it's depended on downstream. There's also a self-reference in the error string at :632 ("Run generage_spectrum_images(...) first.") that should be updated to match.

src/quantem/spectroscopy/dataset3deds.py
```diff
@@ -0,0 +1,3643 @@
...
+
+        energies, _, labels = self.x_ray_lookup(elements)
+        keep = (energies > self.energy_axis.min()) & (energies < self.energy_axis.max())
+        energies = energies[keep]
+        labels = [label for label, ok in zip(labels, keep) if ok]
+
+        mask = (self.energy_axis[:, None] > energies[None, :] - width) & (
+            self.energy_axis[:, None] < energies[None, :] + width
+        )
+        # n, h, w = self.array.shape
```
cophus: This is the old energy-first (n, h, w) implementation left commented out. Since the refactor is settled, safe to delete.

src/quantem/spectroscopy/dataset3deds.py
```diff
@@ -0,0 +1,3643 @@
...
+        # maps = (mask.astype(self.array.dtype).T @ self.array.reshape(n, -1)).reshape(
+        #    mask.shape[1], h, w
+        # )
+
+        w, h, n = self.array.shape
```
cophus: Third naming variant for the scan axes (w, h here vs n_x/n_y elsewhere). For consistency with the proposed convention: scan_row, scan_col, n_energy = self.array.shape, and update the reshape(..., w, h) on :505-506 to match. (Heads-up: the output map is (n_lines, shape[0], shape[1]) — worth a quick sanity check on a non-square scan to confirm row/col land where intended.)

src/quantem/spectroscopy/dataset3deds.py
```diff
@@ -0,0 +1,3643 @@
...
+                "element_names": model.peak_model.element_names,
+                "edge_concentrations": shell_concs,
+                "edge_names": list(model.peak_model.shell_group_names),
+                "edge_element_indices": shell_element_indices,
+                "peak_widths": final_fwhm,
+                "energy_axis": energy_axis.detach().cpu().numpy(),
+                "fit_range": energy_range,
+            }
+
+        n_y, n_x, n_energy = spectra.shape
```
cophus: Same naming nit (n_y, n_x) — fine to leave if you adopt scan_row, scan_col project-wide, just flagging so it's not missed when standardizing. Logic is correct (energy stays last through reshape(n_pixels, n_energy) and view(n_y, n_x, n_elements)).

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        if not self:
+            return "<b>Model Elements:</b><br>&nbsp;&nbsp;None"
+        rows = "".join(
+            f"<tr><td><b>{el}</b></td><td>{', '.join(sorted(info.keys())) if isinstance(info, dict) and info else ''}</td></tr>"
+            for el, info in self.items()
+        )
+        return f"<b>Model Elements:</b><table>{rows}</table>"
+
+
+class Dataset3dspectroscopy(Dataset3d):
```
cophus, line 59: Since energy is now the last axis, the inherited Dataset3d.show() / to_dataset2d() will slice dim0 (a scan axis) as "frames" — e.g. .show() would render scan-row slices rather than something spectroscopy-meaningful. Might be worth overriding show() (or documenting that users should use show_mean_spectrum / show_energy_window_map) so the base behavior isn't surprising.

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        coords = np.column_stack((yy.reshape(-1), xx.reshape(-1)))
+        _, neighbor_indices = cKDTree(coords).query(coords, k=n_neighbors)
+        if n_neighbors == 1:
+            neighbor_indices = neighbor_indices[:, None]
+
+        spectra = array3d.reshape(n_pixels, n_energy)
+        background = np.empty_like(spectra)
+
+        for pixel_index, neighbors in enumerate(neighbor_indices):
+            local_spectrum = spectra[neighbors].mean(axis=2)
```
cophus: _fit_local_background_cube does spectra[neighbors].mean(axis=2), but spectra is 2-D (n_pixels, n_energy), so spectra[neighbors] is (k_neighbors, n_energy) and axis=2 doesn't exist → AxisError. The exception gets re-wrapped as RuntimeError("Background fit failed at pixel ...") at line 1314, hiding the real cause. To average the neighbor spectra into one local spectrum it should be axis=0

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+
+        if show:
+            cmap = kwargs.pop("cmap", "magma")
+            if len(integrated_maps) == 1:
+                selector = next(iter(integrated_maps))
+                self.show_energy_window_map(
+                    energy_window=[energy_min, energy_max],
+                    roi=kwargs.pop("roi", None),
+                    roi_cal=kwargs.pop("roi_cal", None),
+                    mask=selector_masks[selector],
```
cophus, line 560: Integrate passes data_type=... to show_energy_window_map, but that method's signature (dataset3dspectroscopy.py:909) is (energy_window, roi, roi_cal, mask, cmap, show) — no data_type and no **kwargs. So any single-selector call with show=True (both defaults) crashes with got an unexpected keyword argument 'data_type'.

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        """
+        print(f"QuantEM: Aligning {ll.name} and syncing {hl.name}...")
+
+        # 1. Map the drift via argmax
+        zlp_indices = np.argmax(ll.array, axis=2)
+        ref_idx = int(np.median(zlp_indices))
+        shifts = zlp_indices - ref_idx
+
+        # 2. Apply internal QuantEM calibration
+        ll.calibrate_zero_loss_peak()
```
cophus: calibrate_zero_loss_peak() returns a new dataset (line 538, return Dataset3deels.from_array(...)) and does not modify in place, but here its return value is discarded — so "step 2: apply internal QuantEM calibration" does nothing and the function returns ll unaligned. Capture and return the result (e.g. ll = ll.calibrate_zero_loss_peak() and return ll, hl).

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+        if self.attached_spectra is not None:
+            self.attached_spectra.append(two_d_spectrum)
+        else:
+            self.attached_spectra = []
+            self.attached_spectra.append(two_d_spectrum)
+
+    def clear_attached_spectra(self):
+        self.attached_spectra = None
+
+    def plot_attached_spectrum(self, spectrum_index=0):
```
cophus: add_spectrum_to_data stores each entry as a 1-D Dataset1d (intensities only; the energy axis lives in origin/sampling), but plot_attached_spectrum indexes it as a 2×N array ([idx][1] for x, [idx][0] for y). That doesn't yield (energy, intensity), at best it pulls the first two intensity samples, so the plot is wrong. Reconstruct the axis from the stored dataset, e.g.:

```
ds = self.attached_spectra[spectrum_index]
energy = ds.origin[0] + ds.sampling[0] * np.arange(ds.shape[0])
ax_spec.plot(energy, ds.array, linewidth=1.5)
```

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+            ax2.set_title("Thickness Distribution")
+            ax2.set_xlabel(r"$t/\lambda$")
+            ax2.legend()
+
+            plt.tight_layout()
+            plt.show()
+
+        return t_over_lambda
+
+    def interpret_thickness_quality(t_over_lambda, a=0.3, b=1, c=2, dataset=None):
```
cophus: Many methods are missing "self" as an arg.

dataset3deels.py:702, :771, :546, :567, :599, :656, :809, :874, :949

Some of these sort of work, but many are broken

src/quantem/spectroscopy/dataset3dspectroscopy.py
```diff
@@ -0,0 +1,1355 @@
...
+            - 'reconstructed': reconstructed dataset (dataset3dspectroscopy) using n_components
+        """
+
+        from quantem.spectroscopy import Dataset3deds, Dataset3deels
+
+        data = np.asarray(self.array, dtype=float)
+        nx, ny, n_energy = data.shape
+        n_pixels = ny * nx
+
+        spectra = np.moveaxis(data, 2, -1).reshape(n_pixels, n_energy)
```
cophus: dataset3dspectroscopy.py:395, 437, 449: `perform_pca` returns a spatially scrambled reconstruction.

line 395 flattens with np.moveaxis(data, 2, -1) — which is a no-op for a 3-D array (energy is already last) — giving C-order over (shape[0], shape[1]). But the un-flatten at line 449 (reshape(ny, nx, n_energy).transpose(1, 0, 2)) and the loadings reshape at 437 (reshape(n_components, ny, nx)) assume the opposite pixel order. Net effect: the reconstructed cube (and spatial loadings) come out transposed for square scans and scrambled for non-square scans. I'd remove the dead moveaxis and make the un-flatten mirror the flatten exactly.

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+            ``input_spectrum``, ``background_spectrum``, ``concentrations``,
+            ``element_names``, ``peak_widths``, ``energy_axis``, and
+            ``fit_range``.
+        """
+        optimizer_name = str(optimizer).lower()
+        if optimizer_name not in {"adam", "lbfgs"}:
+            raise ValueError("optimizer must be 'lbfgs' or 'adam'")
+
+        if device is None:
+            device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
```
cophus: dataset3deds.py:2808 and :3098 — device selection bypasses quantem.core.config (ignores MPS)

device = torch.device("cuda" if torch.cuda.is_available() else "cpu") hardcodes the choice. The rest of quantem resolves the device through config.get("device") (see diffractive_imaging/ptychography_base.py:603, direct_ptychography.py:518, etc.), and config.validate_device explicitly supports "mps". As written, these fitting routines silently fall back to CPU on Apple Silicon and ignore a user's configured device. Suggest defaulting device to config.get("device").

src/quantem/spectroscopy/dataset3deels.py
```diff
@@ -0,0 +1,989 @@
...
+        # Mask energy window for peak finding
+        mask = (energies > search_window[0]) & (energies < search_window[1])
+        search_energies = energies[mask]
+
+        # Calculate peak map and absolute deviation
+        peak_indices = np.argmax(data[:, :, mask], axis=2)
+        zlp_map_ev = search_energies[peak_indices]
+        absolute_shift = np.abs(zlp_map_ev)
+
+        # Visualization
```
cophus: dataset3deels.py plotting (:587, :682, :757, :832, :935, …) all use raw matplotlib instead of show_2d

These plot_* methods build spatial maps with plt.imshow/plt.colorbar directly, while the convention (and the parent class, e.g. show_mean_spectrum) uses quantem.core.visualization.show_2d with a ScalebarConfig. The file even imports show_2d (line 11) but only uses it once. Routing these through show_2d would give consistent styling and calibrated scalebars.

src/quantem/core/datastructures/dataset1d.py
```diff
@@ -0,0 +1,216 @@
...
+        Returns
+        -------
+        Dataset1d
+
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from quantem.core.datastructures import Dataset1d
+        >>> arr = np.random.rand(10)
+        >>> data = Dataset3d.from_array(arr)
```
cophus: dataset1d.py:99, 143-146, 186-216, copy-paste docstring errors in the new Dataset1d

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+            mean_background_spectrum_all = bg_final.mean(dim=0).cpu().numpy()
+
+            abundance_maps = conc_final.view(n_y, n_x, n_elements).permute(2, 0, 1).cpu().numpy()
+            peak_widths = nn.functional.softplus(peak_width_params).detach().cpu().numpy()
+
+        pytorch_spectrum_images = self._build_pytorch_spectrum_images(
+            abundance_maps=abundance_maps,
+            element_names=list(global_model.peak_model.element_names),
+        )
+        if hasattr(self, "_spectrum_images_pytorch"):
```
cophus: dataset3deds.py:3431-3441 — dead if/else: the else sets {} then immediately rebuilds from {}, so both branches do the same thing. Collapses to:

```
base = getattr(self, "_spectrum_images_pytorch", {})
self._spectrum_images_pytorch = {**base, **pytorch_spectrum_images}
```

src/quantem/spectroscopy/dataset3dxeds.py
```diff
@@ -0,0 +1,3643 @@
...
+        return_maps : bool, optional
+            If ``True``, return ``(maps, labels)``.
+
+        Returns
+        -------
+        tuple[ndarray, list[str]] | None
+            Only returned when *return_maps* is ``True``.
+        """
+        if elements is None:
+            if self.model_elements is None:
```
cophus: dataset3deds.py:487, if self.model_elements is None: never fires (it's initialized to an empty _ModelElementsDict, never None), so the "elements must be specified" guard is dead and users hit a confusing downstream error instead. Use if not self.model_elements: (matches the sibling checks at :2815, :3091).

src/quantem/spectroscopy/spectroscopy_models.py
```diff
@@ -0,0 +1,377 @@
...
+        )
+
+    def forward(self):
+        background = torch.zeros_like(self.energy_axis)
+        for i, coeff in enumerate(self.coeffs):
+            background += coeff * (self.energy_norm**i)
+        return background
+
+
+class ExponentialBackground(nn.Module):
```
cophus: ExponentialBackground is never used anywhere, and its decay acts on the raw (un-normalized) energy axis unlike PolynomialBackground. Remove it, or wire it up consistently.

src/quantem/spectroscopy/utils.py
```diff
@@ -0,0 +1,110 @@
...
+                continue
+
+            energy_kev = _parse_float(row, ("energy_keV", "energy (keV)", "energy"))
+            if energy_kev is None:
+                energy_ev = _parse_float(row, ("energy_eV", "energy (eV)"))
+                if energy_ev is None:
+                    continue
+                energy_kev = energy_ev / 1000.0
+
+            # Use the normalized CSV column as the X-ray line weight.
```
cophus: _parse_float(row, ("col4_norm", "weight", ...)) references a col4_norm column that no shipped CSV has; the "normalized CSV column" comment is misleading. Works only by falling through to weight, drop the stale key.

cophus (commented):
Really nice to see the spectroscopy module come together! I went through the data-layout side carefully and the (E, x, y), (scan, scan, energy) refactor is consistent across both EELS and EDS. Energy is the last axis everywhere, including the torch fitting path. 

One thing that would make it more robust: the two scan axes get three different names across the module (n_x, n_y / n_y, n_x / w, h), and the EELS class docstring says (scan_y, scan_x, energy). The indexing is correct everywhere, so these aren't bugs, but it makes the code hard to read and easy to break later. Could we standardize on one convention?

Proposed: arrays are (scan_row, scan_col, energy); unpack as scan_row, scan_col, n_energy = self.array.shape; loop indices i_row, i_col; spectra as self.array[i_row, i_col, :].

src/quantem/spectroscopy/dataset3dxeds.py
smribet: Yes, I agree! @arthurmccray and I discussed this before. Perhaps after finishing this PR, we can do this next.

src/quantem/spectroscopy/dataset3deels.py
smribet: Great question! I don't think we know yet. @nwvlahakis

nwvlahakis: Most likely some datasets will need more complicated/polynomial fits, right? But let me test this code on a bunch of LL EELS data I have and see if there's any where the plane is clearly insufficient

src/quantem/spectroscopy/dataset3deels.py
nwvlahakis: this function was redundant with measure_zlp_offset and apply_zlp_correction, so it's been removed. Can restore and remove the other redundant functions if that's preferred

cophus (approved):
Looks good!

PR #197, opened 2026-03-19
Adding DOI and contributors for improved citations

arthurmccray:
Creating this as a draft PR so that other folks can see and use this branch. I think the plan is to use a `Zenodo.json` so that each release gets a proper DOI, and maybe also a manually updated `CONTRIBUTORS.md` file? Or there might be a way to get that to auto-populate based on the github users who have commits that made it to dev.


arthurmccray:
@gvarnavi just FYI that I made this--if you'd rather do your own branch and PR it in when done that's fine too and you or I can delete this.

PR #198, opened 2026-03-19
Imaging refactor part 1 - Lattice class - constructor, properties and define_lattice_vectors and plot functions

darshan-mali:
## What does this PR do?

Part 1 of Imaging refactor

Creates the basic Lattice infrastructure for imaging. This includes 
- Constructor that follows the correct API `Lattice.from_data()`
- Lattice class property `Lattice.image : Dataset2d` and its corresponding setter function
- The `Lattice.define_lattice_vectors()` function
- The `plot()` function, `PLOT_REGISTRY` and `plot_lattice_vectors()` function

## API and logic

A `Lattice` object should always be created using the `from_data()` method.
Creating it directly from `Lattice(data)` will raise `RuntimeError: Use Lattice.from_data() to instantiate this class.`
Using `from_data()` requires the data to be either `Dataset2d` or `numpy.ndarray`

The dataset can be accessed by `Lattice.image`. This returns a `Dataset2d` object even if the data used to instantiate was `numpy.ndarray`

The `Lattice.define_lattice_vectors()` function defines the Lattice using the `r = r0 + m*u + n*v` definition, where `u` and `v` are lattice vectors and `r0` is any known lattice point. `r0`, `u` and `v` can be optimized and the optimization can be done in steps (especially for larger datasets)

All plotting should be done using the `Lattice.plot()` function. `Lattice.default_plot` keeps track of which plotting function should be called based on the latest functions called. This can be overridden using the `kind` parameter, which allows access to older plots without any recalculation.

The `Lattice` class inherits from `AutoSerialize` and follows standard quantEM I/O

### New files

- `src/quantem/imaging/lattice.py`: This contains the all the infrastructural, functional and API changes mentioned above.
- `src/quantem/imaging/lattice_visualization.py`: This contains all the plotting functions and helpers that will be required for `Lattice`
- `tests/imaging/test_lattice.py`: Basic pytests for `Lattice` based on expected user behaviour

### Updated files

- `src/quantem/imaging/__init__.py`: Updated imports

## Examples

A testing notebook with examples can be found here: 
[Lattice_refactor_1.ipynb](https://github.com/user-attachments/files/26474480/Lattice_refactor_1.ipynb)

Example code block:
```
lat.define_lattice_vectors(
    origin = (1626, 1024),
    v = ( 25 ,  0),
    u = ( 0,   25),
    refine_lattice = True,  # Default = True 
    block_size = 5,
).plot(
    figsize = (8,8)
)
```

Example output:
image

## PR Checklist

- [X] This PR introduces a public-facing change (e.g., API, CLI input/output).
    - [X] For functional and algorithmic changes, tests are written or updated.
    - [X] Documentation (e.g., tutorials, examples, README) has been updated.

## Reviewer checklist

- [ ] The notebook provided runs as expected and no bugs were found


src/quantem/imaging/lattice_visualization.py
```diff
@@ -0,0 +1,143 @@
...
+    bound_num_vectors : int | None
+        Number of lattice vectors to draw in each direction.
+        None clips lines to image edges.
+    **kwargs forwarded to show_2d (e.g. cmap, title).
+    """
+    fig, ax = show_2d(lattice._image.array, returnfig=True, **kwargs)
+    if ax.images:
+        ax.images[-1].set_zorder(0)
+    H, W = lattice._image.shape
+    r0, u, v = (np.asarray(x, dtype=float) for x in lattice._lat)
```
wwmills, line 58: Not sure if the code in its current state could call this function without the _lat attribute existing, but a check for its existence at the beginning of this function and an error if the attribute is not present would be a good catch.

src/quantem/imaging/lattice_visualization.py
```diff
@@ -0,0 +1,143 @@
...
+
+    ax.scatter(
+        r0[1], r0[0], s=60, edgecolor=(0, 0, 0), facecolor=(0, 0.5, 0), marker="s", zorder=30
```
wwmills: the zorder being 30 here and 20 later seems a little arbitrary, could we get this to a more reasonable value? A user might try to plot other things on this axes object and they might have trouble with the z order. At least we could make them adjacent.

src/quantem/imaging/lattice_visualization.py
```diff
@@ -0,0 +1,143 @@
...
+    n_vec = int(bound_num_vectors) if bound_num_vectors is not None else 1
+    for k in range(1, n_vec + 1):
+        tip = r0 + k * u
+        ax.arrow(
+            r0[1],
+            r0[0],
+            (tip - r0)[1],
+            (tip - r0)[0],
+            length_includes_head=True,
+            head_width=4.0,
```
wwmills, line 73: here we have some set widths, which seems okay, but if we do that, maybe we want to give this plotting a default axsize so that the lines show up nice?

src/quantem/imaging/lattice_visualization.py
```diff
@@ -0,0 +1,143 @@
...
+    if ax.images:
+        ax.images[-1].set_zorder(0)
+    H, W = lattice._image.shape
+    r0, u, v = (np.asarray(x, dtype=float) for x in lattice._lat)
+
+    ax.scatter(
+        r0[1], r0[0], s=60, edgecolor=(0, 0, 0), facecolor=(0, 0.5, 0), marker="s", zorder=30
+    )
+
+    n_vec = int(bound_num_vectors) if bound_num_vectors is not None else 1
```
wwmills: in the function header, isn't it requiring that the bound_num_vectors variable is int or None? Is casting as integer necessary here? If so, should we round and cast as int instead of just casting as int (which should always round down)? I don't really care too much about this.

tests/imaging/test_lattice.py
```diff
@@ -0,0 +1,160 @@
...
+
+    def test_init_and_constructor(self):
+        """Test that direct init is blocked and from_data works."""
+        image = np.random.randn(100, 100)
+        ds2d = Dataset2d.from_array(image)
+
+        with pytest.raises(RuntimeError, match="Use Lattice.from_data"):
+            Lattice(ds2d)
+
+        lattice = Lattice.from_data(image)
```
wwmills: should also test that from_data will work with a Dataset2d object

tests/imaging/test_lattice.py
```diff
@@ -0,0 +1,160 @@
...
+        """Test min/max normalization."""
+        image = np.random.randn(100, 100) * 1000.0
+        image[0, 0] = 0.0
+        image[99, 99] = 1000.0
+
+        # Both normalizations
+        lattice = Lattice.from_data(image)
+        assert lattice.image.array.min() == 0
+        assert lattice.image.array.max() == 1
+
```
wwmills, line 38: could test just min and just max normalization

tests/imaging/test_lattice.py
```diff
@@ -0,0 +1,160 @@
...
+        lattice = Lattice.from_data(image)
+
+        # Wrong shape
+        with pytest.raises(ValueError):
+            lattice.define_lattice_vectors(origin=[1, 2, 3], u=[5, 0], v=[0, 5])
+
+        # Negative block_size
+        with pytest.raises(ValueError):
+            lattice.define_lattice_vectors(origin=[50, 50], u=[5, 0], v=[0, 5], block_size=-1)
+
```
wwmills, line 144: how about lattice vectors that are identical, or origins that are outside of the image? Could add a catch for that in the lattice file

wwmills (changes requested):
Looks good to me, just a few minor suggestions

src/quantem/imaging/lattice_visualization.py
darshan-mali: Added an additional check for it.

src/quantem/imaging/lattice_visualization.py
darshan-mali: Set the zorders to 4 and 5.

src/quantem/imaging/lattice_visualization.py
darshan-mali: Set the default figsize = (10,10)

tests/imaging/test_lattice.py
darshan-mali: Added checkers for this in the main file, along with additional conditions in the pytest. 

Enforced invertability of lattice vectors, which checks for more conditions in addition to whether they are identical or not.

darshan-mali:
@wwmills, I have added all the changes you requested. It is ready for review again.

tests/imaging/test_lattice.py
darshan-mali: Added these conditions to the test

tests/imaging/test_lattice.py
darshan-mali: Added these conditions to the test.

src/quantem/imaging/lattice_visualization.py
darshan-mali: So, apparently, Python does not enforce typehints. They are only used by the IDE, linters, etc., but not by the Python interpreter itself. So, type checking needs to be enforced manually.

I added np.ceil() to avoid direct casting. I think ceil would be better in this case over round.

wwmills (approved):
My comments have been addressed

PR #207, opened 2026-04-06
fix for cropping

smribet:
Previously when cropping a dataset, the origin wasn't shifted, which led to erroneous calibrations for spectroscopy data, for example.

### DIFF
--- a/src/quantem/core/datastructures/dataset.py
+++ b/src/quantem/core/datastructures/dataset.py
@@ -23,7 +23,7 @@ class Dataset(AutoSerialize):
     Attributes (Properties):
         array (NDArray): The underlying n-dimensional NumPy array data.
         name (str): A descriptive name for the dataset.
-        origin (NDArray): The origin coordinates for each dimension (1D array).
+        origin (NDArray): The origin coordinates for each dimension (1D array) in calibrated units.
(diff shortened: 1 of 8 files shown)


arthurmccray:
This looks good, my two main comments:

- Are the units of `origin` pixels or `.units` units? I can't find in any of our documentation which it should be, but we need to be explicit either way. I believe @cophus had sort of implemented an `origin_units` property, but it hasn't made it into dev yet and idk if it ever will. It is languishing in an old commit on the `fitting_models` branch, but that branch will never be merged cuz it has commits (like this one) that aren't relevant, and we've moved to `fitting_models_clean`. 
    - image

- we should probably also update slicing `Dataset` as well. Currently `dset2 = dset[a:b]` is roughly equivalent to `dset2 = dset.crop((a, b), axes=0)`, so we should also probably update the `__getitem__` method so it updates the origin also.

smribet:
I believe the origin is calibrated.

arthurmccray:
> I believe the origin is calibrated.

great! do you mind updating the docstrings for `Dataset` to make it explicit?

smribet:
Yes of course!

arthurmccray (approved):
lgtm!

PR #210, opened 2026-04-15
diffractive_imaging -> dev for OptimizerParams and similar

arthurmccray:
a few small changes so that ptycho works with `OptimizerParams`, most of the commits are just merges, only file changes are in diffractive_imaging except for one docstring.


cedriclim1 (approved):
LGTM

PR #213, opened 2026-04-24
Per-parameter Learning Rate (PPLR) Implementation, Tensor Decomposition Models, and Tomography Updates

cedriclim1:
# What does this PR do?

First pass implementation of `KPlanes` implementation (https://brentyi.github.io/tilted/, https://sarafridov.github.io/K-Planes/). These tensor decomposition models require multiple parmeter optimization which is not handled in `OptimizerMixin`.

The solution @arthurmccray and I came up with is to setup a per-parameter learning rate `PPLR` abstract class which will be instantiated with models that have multiple parameters to be optimized to. This is implemented in`core/ml/models/model_base.py`. The most important part about this class is the `get_params` method, which is used for parsing the trainable parameters (see `ObjectTensorDecomp`'s `get_optimization_parameters`) .

```
class PPLR(ABC):
    """
    Abstract base class for models that require multi-scale parameter optimization.
    """

    @abstractmethod
    def get_params(self) -> Dict[str, list[nn.Parameter]]:
        """
        Return a dictionary of parameters grouped by key.

        For example if your nn.Module has multiple optimizable parameter groups,
        you can return a dictionary with the keys "grids" and "sigma_net"
        (KPlanes example).
        """
        pass
```

The KPlanes and TILTED implementations are in `core/ml/models` in `kplanes.py` with `KPlanes`, `KPlanesTILTED`, and `CPTilted`. Note `CPTIlted` is used for the two-phase warmup that the paper recommends by pretraining the SO3 rotations of the planes in a lower representation space. `so3params.py` also has both quaternion and R9+SVD implementations which can be swapped as a parameter in `KPlanesTILTED`. 

In `tomography/object_models.py` there is now a new object that handles the new tensor decomposition methods, `ObjectTensorDecomp` that inherits from `ObjectINR` and the following overloads are needed:

- `optimizer_params` property: The setter is changed, since we're building a a dictionary of parameters into our optimizers per https://docs.pytorch.org/docs/stable/optim.html#per-parameter-options.
- `reconnect_optimizer_to_parameters`: This one is the most annoying thing to implement, since now our parameters is a dictionary. 

There are a few optional overrides that I've also included preemptively just incase if I want to add additional soft/hard constraints to these methods as you can see both `apply_soft_constraints` and `apply_hard_constraints` are still there. 

**Key change to OptimizerMixin**: I have changed `set_optimizer` to handle a dictionary of parameters. I think this is better than my original idea of just implementing this in `object_models.py` since you would have to overload that function again. I think this is fine and still works for single optimization stuff.

There are also small changes to different aspects of the Tomography module, but mostly contained in `object_models.py`. Listing out the changes and rationale in the tomography files:

- `tomography.py`: I've disabled `torch.bfloat16` autocasting since `F.grid_sample` does not have `autograd` implemented errors. I've also put the TILTED tensor decomp pretraining check by looking at the convergence of the SO3 rotations. I don't like this being here, but I'm not quite sure where else to put this.
- `tomography_base.py`: Type-hinting change for setting up distributed.
- `tomography_opt.py`: Parsing is already handled in `object_models.py`. 

# What the reviewer should do
Attached is a notebook that tests tensor decomposition methods in tomography: [0415_PPLR_testing.ipynb](https://github.com/user-attachments/files/27032105/0415_PPLR_testing.ipynb). The dataset here is from `quantem-tutorials`, so just set the directory of the same dataset to this notebook. The runtime can be increased by increasing the batch size, but the learning rates have to be scaled. **Setting the batch size to 4096 seems to be fine? So feel free to change it and scale the learning rates by 2**. 

- [x] Double check implementations across the board, especially the `optimizer_mixin.py`, and `object_models.py`
- [x] Organization makes sense for tensor decomposition stuff. I'm actually not very happy with how `core/ml` is looking. Where do we draw the line of what lives in models (I think everything model related should live in there tbh: INR, CNN, etc...)? 
- [x] Check reconstruction workflows are still working i.e, ptychography due to the OptimizerMixin change.


.vscode/settings.json
```diff
@@ -78,4 +78,5 @@
         "yticks"
     ],
     "basedpyright.analysis.typeCheckingMode": "standard",
+    "python.REPL.enableREPLSmartSend": false,
```
arthurmccray: what's this for?

bobleesj:
@cedriclim1 the default VS code setting has been removed, so please pull from `upstreadm/dev`

src/quantem/core/ml/constraints.py
```diff
@@ -7,6 +7,8 @@
 import torch
 from numpy.typing import NDArray
 
+from quantem.tomography.tomography_context import ReconstructionContext
```
arthurmccray: If we're going to have the `ReconstructionContext` used in our `Constraints` `ABC`, then it shouldn't live in `tomography_context`. 

It's definitely not going to be fully backwards compatible, but hopefully things should run without errors, and we'll only have problems with linting?

src/quantem/core/ml/models/model_base.py
```diff
@@ -0,0 +1,50 @@
+from abc import ABC, abstractmethod
+from typing import Dict
+
+import torch.nn as nn
+
+
+class PPLR(ABC):
+    """
+    Abstract base class for models that require multi-scale parameter optimization.
```
arthurmccray: is "multi-scale" the right way of describing this?

src/quantem/core/ml/models/model_base.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Planar factored-grid models: K-Planes, K-Planes-TILTED, HexPlane, tri-planes.
+
+    Subclasses must set ``grids``, ``tilted``, and ``resolution`` as normal
+    attributes in ``__init__``.
+    """
+
+    grids: nn.ParameterList
+    tilted: bool
+    resolution: list[int]
```
arthurmccray, line 50: what's the motivation for having the `TensorDecomposisionModel` base class, shouldn't there just be a `td_type` in `PlanarDecompositionModel`? Unless there's going to be other, non-planar decomp moels...

src/quantem/core/ml/models/so3params.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def extra_repr(self) -> str:
+        return f"T={self.quats.shape[0]}"
+
+
+class SO3ParamR9SVD(nn.Module):
+    """
+    SO(3) rotation bank using R9+SVD parameterization.
+    Each rotation is stored as an unconstrained 3x3 matrix M,
+    projected to SO(3) via SVD+(M) = U diag(1,1,det(UVt)) Vt.
```
arthurmccray, line 153: I would add a link to the paper in the docstring

src/quantem/core/ml/models/so3params.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+
+    def __init__(self, T: int, init: str = "random"):
```
arthurmccray: `inti: Literal['random', 'identity']`

src/quantem/core/ml/optimizer_mixin.py
```diff
@@ -606,25 +626,44 @@ def set_optimizer(self, opt_params: OptimizerType | dict | None = None) -> None:
...
-            p.requires_grad_(True)
+        for group in params:
+            for p in group["params"]:
+                p.requires_grad_(True)
+        # Figure out which optimizer class to use
+        if isinstance(self._optimizer_params, dict):
+            # Per-group case: all groups must agree on the optimizer class,
+            # and per-group hyperparameters are already baked into each dict
+            # by get_optimization_parameters().
+            opt_specs = list(self._optimizer_params.values())
```
arthurmccray: is there a reason you use the hidden `self._optimizer_params` rather than `self.optimizer_params`? Not that it makes a difference, just curious.

src/quantem/core/ml/optimizer_mixin.py
```diff
@@ -551,18 +553,36 @@ def scheduler(self) -> "torch.optim.lr_scheduler.LRScheduler | None":
         return self._scheduler
 
     @property
-    def optimizer_params(self) -> OptimizerType:
+    def optimizer_params(self) -> OptimizerType | dict[str, OptimizerType]:
```
arthurmccray: if we're making everything consistent, should we just have `optimizer_params` always be a `dict[OptimizerType]`?

This would also make `get_optimization_parameters` and other methods simpler as they wouldn't have to handle both cases

src/quantem/core/ml/models/so3params.py
```diff
@@ -0,0 +1,138 @@
...
+            dim=-1,
+        ).reshape(-1, 3, 3)  # (T, 3, 3)
+
+        return R
+
+    def extra_repr(self) -> str:
+        return f"T={self.quats.shape[0]}"
+
+
+class SO3ParamR9SVD(nn.Module):
```
arthurmccray, line 149: should probably add a classmethod `from_matrix` so one could initialize it with a value that's close. Or actually, have a static method that goes matrix -> R9 and then just call that with a classmethod.

src/quantem/core/visualization/visualization.py
arthurmccray, line 1: no idea why this had different formatting previously, but it doesn't matter

src/quantem/tomography/dataset_models.py
```diff
@@ -643,6 +648,32 @@ def to(self, device: torch.device | str):
...
+        )
+
+    def load_parameters(self, path: str):
+        """
+        Loads the learned parameters from a file.
+        """
+        data = torch.load(path)
+        self._z1_params = nn.Parameter(data["z1"])
+        self._z3_params = nn.Parameter(data["z3"])
+        self._shifts_params = nn.Parameter(data["shifts"])
```
arthurmccray: should this put them on `self.device`?

src/quantem/tomography/tomography.py
```diff
@@ -235,6 +245,25 @@ def reconstruct(
                 total_loss += batch_loss.detach()
                 consistency_loss += batch_consistency_loss.detach()
 
+            if isinstance(self.obj_model.model, CPTilted):
```
arthurmccray, line 248: This could query `model.tilted` or similar, right?

src/quantem/tomography/tomography_context.py
```diff
@@ -0,0 +1,21 @@
...
+import torch
+
+
+@dataclass
+class ReconstructionContext:
+    """
+    Handles all reconstruction parameters to be passed into object models.
+
+    Subclasses will pick whatever parameter they need
+        - Pixelated reads ".volume"
```
arthurmccray, line 14: Should `volume` be initialized as well? I would include in the docstring what the actual things are (e.g. what is `all_densities`?) Presumably only the relevant ones are initialized and the others left as None?

src/quantem/__init__.py
arthurmccray, line 1: okay the fact that your formatter is changing this file is a little suspicious, are you using a weird ruff setup or anything?

src/quantem/core/ml/models/kplanes.py
arthurmccray, line 1: cite the relevant papers please for the models

src/quantem/tomography/object_models.py
```diff
@@ -808,43 +852,203 @@ def create_volume(self, return_vol: bool = False):
...
-    def to(self, device: str | torch.device):  # pyright: ignore[reportIncompatibleMethodOverride] -> better to do this device change
-        if isinstance(device, str):
-            device = torch.device(device)
-        self._device = device
-        if self.world_size == 1:
-            self._model = self._model.to(device)
-        elif not isinstance(self._model, torch.nn.parallel.DistributedDataParallel):
-            self.distribute_model(self._model)
-        self.reconnect_optimizer_to_parameters()
+    def _get_plane_tv_loss(self) -> torch.Tensor:
```
arthurmccray, line 946: some explanation for what this is would be helpful.  I guess also why you use this _and_ the normal INR style volume TV.

src/quantem/tomography/object_models.py
```diff
@@ -153,15 +184,26 @@ def parse_dict(
...
 
 ObjConstraintsType = (
-    ObjConstraintParams.ObjPixelatedConstraints | ObjConstraintParams.ObjINRConstraints
+    ObjConstraintParams.ObjPixelatedConstraints
+    | ObjConstraintParams.ObjINRConstraints
+    | ObjConstraintParams.ObjTensorDecompConstraints
 )
 
 
+def _unwrap(model: nn.Module | nn.parallel.DistributedDataParallel) -> PlanarDecompositionModel:
```
arthurmccray, line 200: I'm seeing linter errors for this function and in a couple other places. Not a huge deal but wanted to let you know

src/quantem/tomography/object_models.py
arthurmccray, line 1: just did a pretty quick skim over this and it looks good. Main thing is that, since this will be new to most folks, include some resources and more explanation for what's going on. i.e. reference the relevant papers and more code comments or docstrings

arthurmccray (commented):
overall looks good! I made the necessary few changes to ptycho so it works with this, and overall i'm happy with it. Hopefully we're one step closer to not having to mess with `OptimizerMixin` ever again...

As to `/core/ml` organization, yes `cnn.py`, `cnn_dense.py`, `inr.py`, etc. should all be in `/ml/models`. I don't think it should be that bad to move them actually? An automatic refactor should handle everything in `quantem` fine, and they'll be accessible at the same level of the namespace (we still want to be able to `from quantem.core.ml import CNN2d`), so it shouldn't affeect the tutorials either really. If you move them (and whatever else you think) we can just re-run all the tutorials and tests to make sure nothing breaks.

.vscode/settings.json
cedriclim1: This is outdated re [Bob](https://github.com/electronmicroscopy/quantem/pull/213#issuecomment-4331396571)'s comment

src/quantem/core/ml/constraints.py
cedriclim1: I don't know what I was thinking here. See commit [adcc26d](https://github.com/electronmicroscopy/quantem/pull/213/commits/adcc26dc018230d67641c0676d71b4cb877b4c2a). I think this is the most non-invasive thing to do.

src/quantem/core/ml/models/model_base.py
cedriclim1: Description changed to "multi-parameter optimization". [03da610](https://github.com/electronmicroscopy/quantem/pull/213/commits/03da610c5a3ce34c601058bfb243ba1d4d8ba42c)

src/quantem/core/ml/models/model_base.py
cedriclim1: Yes, this is just inplace just in case we want to implement other Tensor decomposition as dicussed in this paper: https://arxiv.org/pdf/2503.21507. I might implement them in the future, but also maybe not...This is just here in the case I do want to implement (CP, TT, and TU decompositions).

src/quantem/core/ml/models/so3params.py
cedriclim1: Done

src/quantem/core/ml/models/so3params.py
cedriclim1: Fixed! Thanks

src/quantem/core/ml/optimizer_mixin.py
cedriclim1: I thought it was just good code design to have a hidden variable and define the property + setter? Feel free to correct me, but it also allows me to have this `_normalize_optimizer_params` to standardize the output , that I guess can be put into the setter. Let me know what you think.

src/quantem/core/ml/optimizer_mixin.py
cedriclim1: Hmmm....I think this is fair, are you thinking `def _normalize_optimizer_params` will always return `dict[str, OptimizerType]`? Not sure if this will break stuff downstream in Ptycho.

src/quantem/tomography/dataset_models.py
cedriclim1: You are correct, good catch.

src/quantem/tomography/tomography.py
cedriclim1: This is more of a debugging step for me and just see if the rotations are converging in the pre-training scheme for the plane rotations using a lower-dimensional representation of the model rather than going straight into K-Planes. I could get rid of this, or add it to the verbose output, i.e., part of the pbar update. Or I could also completely get rid of this.

cedriclim1: I guess this is just good for pretraining too. I just haven't thought of a good solution for using `CPTilted` as a pretraining step for the full `KPlanesTilted` since it does train on the actual dataset, full INR style. It's not learning a SIRT object or anything.

src/quantem/tomography/tomography_context.py
cedriclim1: Yea good catch, I will leave `TomographyConventional` alone for now and integrate a lot of the `FastTomo` stuff in a future PR.

src/quantem/__init__.py
cedriclim1: I'm not sure lol. It seems like it just added extra whitespace to this file for some reason.

src/quantem/core/ml/models/kplanes.py
cedriclim1: Added citations.

src/quantem/tomography/object_models.py
cedriclim1: Added explanation.

cedriclim1:
@arthurmccray ready for review again - addressed most comments, but some need follow-up.

src/quantem/core/ml/models/model_base.py
arthurmccray: good enough for me

src/quantem/core/ml/models/so3params.py
arthurmccray: bump. slight preference to having static methods `rotmat_to_r9` and `r9_to_rotmat` (the latter of which is simply called by `as_matrix`), but just having it in a classmethod would be fine too.

src/quantem/tomography/tomography_context.py
arthurmccray: please include here what these actually mean tho. Like `volume` is pretty self explanatory (presumably it's the full reconstructed volume), and `coords` would be for INR (is it the coords for the full volume? or just part of it), but `pred` and `all_densities` and `obj` should be explained in a line

src/quantem/core/ml/optimizer_mixin.py
arthurmccray: it probably would, but it should be easy for me to fix... but on the other hand it feels a little silly for it to be `{"params": OptimizerType}` for single Optimizers. Up to you!

src/quantem/core/ml/optimizer_mixin.py
arthurmccray: in this case i don't think it makes any difference as the `optimizer_type` property just directly returns `_optimizer_type`. But in theory it might not, and that's more why I think about using the properties internally. idk if that's officially good or bad practice though--in my comment i was honestly asking it as a question 😅

src/quantem/tomography/tomography.py
arthurmccray: Works for me

arthurmccray (commented):
Overall looks good, just one or two things I saw. Did I miss anything that you needed my followup on?

src/quantem/core/ml/models/so3params.py
cedriclim1: Fixed here [990f35d](https://github.com/electronmicroscopy/quantem/pull/213/commits/990f35d27c674b51dfc3e0ce8bfacadc236efbc8)

src/quantem/core/ml/optimizer_mixin.py
cedriclim1: Oh I see what you mean - even if I defined the properties, and why I didn't use it lol. Fixed in [c56994a](https://github.com/electronmicroscopy/quantem/pull/213/commits/c56994a529269d7fc713ee32e0859961f9a6f642).

src/quantem/tomography/tomography_context.py
cedriclim1: Done: [4b8ca47](https://github.com/electronmicroscopy/quantem/pull/213/commits/4b8ca472a0d6fc99980c60d50739782870250e0a)

src/quantem/core/ml/optimizer_mixin.py
cedriclim1: Resolved here: [21944c4](https://github.com/electronmicroscopy/quantem/pull/213/commits/21944c406c804d565d37c956bd7b69d8faa34dde), let me know if this is what you want.

src/quantem/core/ml/optimizer_mixin.py
cedriclim1, line 1: Line 222: OptimizerType --> OptimizerParamsType

arthurmccray:
@cedriclim1 I fixed the `OptimizerParamsType` and changed `SchedulerParamsType` to match. Want you to check once but then I'll approve and merge

arthurmccray (approved):
noice. finally

Issue #124, opened 2025-11-05
Consider `"python.languageServer": "pylance",` in `setting.json` so that VS Code can  detect functions/classes in quantem after `pip install -e .`


bobleesj:
I will bring this up in the upcoming dev's meeting.  cc @smribet 

Basically change from `"python.languageServer": "None",` to `"python.languageServer": "Pylance",` in `settings.json`.  This enables my VC Code to auto-detect functions and classes.

bobleesj:
closed by the above PR.

PR #214, opened 2026-04-24
Delete `.vscode` directory to allow devs to maintain their own custom VS code settings

bobleesj:
### What problem PR addreses

This `settings.json` often enforces and overrides my default setup. Hence I no longer have VS code type inferred during dev and  in notebooks while checking out other dev's PR or branches. 

This was initially addressed in https://github.com/electronmicroscopy/quantem/issues/124 but a better solution is to delete this file and let the dev decide their own env and path - whether uv/virutal-env/conda/etc. 

### What should reviewer(s) do

Review and merge please if appropriate.


bobleesj:
@arthurmccray this is ready for review

arthurmccray (approved):
makes sense to me

bobleesj:
thanks for fast review!

PR #225, opened 2026-05-15
`Dataset4dstemTorch` - simple data class to containerize VRAM 4DSTEM data for streamlined 4DSTEM workflows

### LINKED ISSUE #222: refactor: dataset accepts torch (cpu/gpu) array-like for streamlined 4DSTEM algo & workflows
### Problem                                                                                                                  
                                                        
The main goal of the Dataset class is to containerize and retrieve the underlying data, including metadata.              

Currently it is forced to `numpy`. This deviates from the main workflow for 4D-STEM reconstructions, which primarily rely  
on `torch` or `cupy` - both for using `quantem` directly and for algo dev in forks.
                                                                                                                        
Ref:                                                   

```python
if not isinstance(arr, np.ndarray):
    raise TypeError(                                                                                                     
        "Dataset requires a NumPy array (CuPy is not supported on this branch)."
    )                                                                                                                    
```

Two cases where I don't get to use `Dataset4dstem`:                                                                       

Case i) When a 4D-STEM dataset is loaded via the CUDA path (https://github.com/electronmicroscopy/quantem/pull/135),     
wrapping it in `Dataset4dstem` requires converting from VRAM to CPU RAM. With a duck-typed dataset, the data can stay in
VRAM and the widget (`Show4DSTEM`) reads it directly, eliminating the RAM↔VRAM round trip.                                 
                                                        
Case ii) Live SSB. Detector frames stream into VRAM and are processed on the GPU - they never touch CPU RAM.             

### Proposed solution                                                                                                        
                                                        
Remove `isinstance(arr, np.ndarray)` and replace it with a minimal duck-type check.

Accept any object exposing shape, ndim, dtype, and the standard reduction methods (.sum, .mean, .max, .min).

bobleesj:

Closes https://github.com/electronmicroscopy/quantem/issues/222

Simple solution to use native `quantem` datastructure for all downstream 4DSTEM workflows:

Effect: avoid VRAM-RAM-VRAM copy, less memory footprint, faster loading for widgets and vectorization operations.

Demo:

sepen_gif_20260514_190238

No side effect on any of the existing codebase. Made this class dead simple - add more features as needed. 

It is primarily used by my code and will later extend to 5DSTEM (tilt/time).

### DIFF
--- a/src/quantem/core/datastructures/dataset4dstem_torch.py
+++ b/src/quantem/core/datastructures/dataset4dstem_torch.py
@@ -11,11 +11,21 @@ class Dataset4dstemTorch:
     (``array``, ``name``, ``origin``, ``sampling``, ``units``,
     ``signal_units``).
 
-    Goal: keep 4D-STEM data on the GPU end-to-end and skip VRAM <-> RAM
-    round-trips. Each hop is expensive on multi-GB datasets and doubles
-    peak memory. Wrapping a cupy or torch GPU array here is effectively
-    free and holds the original VRAM allocation.
+    ``Dataset4dstem`` wraps a numpy array and lives in CPU RAM. Use it
+    when the raw data starts on the host (file readers that return numpy,
+    CPU-only analysis).
+
+    Use ``Dataset4dstemTorch`` instead when the raw data already lives on
+    the GPU - any CUDA pipeline producing torch / cupy arrays, live
+    streaming detector frames, GPU file readers. Wrapping the existing
+    GPU array is effectively free and the data stays in VRAM end-to-end.
+    Going through the CPU class instead forces a copy from GPU to CPU on
+    wrap and another copy from CPU back to GPU when the consumer
+    re-uploads, which is expensive on multi-GB datasets and doubles peak
+    memory.
     """
+
+    _token = object()
 
     def __init__(
         self,
@@ -25,7 +35,10 @@ class Dataset4dstemTorch:
         sampling: tuple[float, ...] | list[float] | None = None,
         units: list[str] | tuple[str, ...] | None = None,
         signal_units: str = "arb. units",
+        _token: object | None = None,
     ):
+        if _token is not self._token:
+            raise RuntimeError("Use Dataset4dstemTorch.from_array() to instantiate this class.")
         if not (isinstance(array, torch.Tensor) and array.ndim == 4):
             raise TypeError("Dataset4dstemTorch requires a 4D torch tensor")
         self.array = array
@@ -59,4 +72,5 @@ class Dataset4dstemTorch:
             sampling=sampling,
             units=units,
             signal_units=signal_units,
+            _token=cls._token,
         )
--- a/src/quantem/core/datastructures/dataset4dstem_torch.py
+++ b/src/quantem/core/datastructures/dataset4dstem_torch.py
@@ -32,6 +32,7 @@ class Dataset4dstemTorch:
         sampling: tuple[float, ...] | list[float] | None = None,
         units: list[str] | tuple[str, ...] | None = None,
         signal_units: str = "arb. units",
+        metadata: dict | None = None,
         _token: object | None = None,
     ):
         """Initialize a torch-backed 4D-STEM dataset.
@@ -51,6 +52,10 @@ class Dataset4dstemTorch:
             Units for each dimension. If None, defaults to ["pixels"] * 4.
         signal_units : str, optional
             Units for the array values, by default "arb. units".
+        metadata : dict | None, optional
+            Mirror of ``Dataset4dstem.metadata``. Auto-populates
+            ``r_to_q_rotation_cw_deg`` and ``ellipticity`` keys with
+            ``None`` so consumers expecting them do not need a guard.
         _token : object | None, optional
             Token to prevent direct instantiation, by default None.
         """
@@ -76,6 +81,9 @@ class Dataset4dstemTorch:
         self.sampling = np.asarray(sampling if sampling is not None else (1, 1, 1, 1), dtype=float)
         self.units = list(units) if units is not None else ["pixels"] * 4
         self.signal_units = signal_units
+        self.metadata = dict(metadata) if metadata is not None else {}
+        for k in ("r_to_q_rotation_cw_deg", "ellipticity"):
+            self.metadata.setdefault(k, None)
 
     @classmethod
     def from_array(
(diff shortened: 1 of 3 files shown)


bobleesj:
@cedriclim1 thanks for your time and review!

src/quantem/core/datastructures/dataset4dstem_torch.py
```diff
@@ -0,0 +1,152 @@
...
+    transfer on wrap and a host-to-device transfer on consume, doubling
+    peak memory.
+
+    Notes
+    -----
+    Construction is gated by a token: use ``Dataset4dstemTorch.from_array``
+    rather than calling the constructor directly.
+    """
+
+    _token = object()
```
cedriclim1, line 25: You've initialized a `_token` here, is there any reason why you're not inheritting from AutoSerialize?

bobleesj (commented):
@cedriclim1 

autoserialize added. 

if I understand correctly, `token` is a fancy technique to prevent direct init of data class

src/quantem/core/datastructures/dataset4dstem_torch.py
bobleesj: done -autoserialize added

cedriclim1 (commented):
Ergh, I'm not very happy that this has to be implemented. This kind of goes against the design principles Arthur has thought about I think i.e., they should be compatible with AutoSerialize, and most likely more things that I'm forgetting. But, I can see why you did this since everything in `core/datastructures` is implemented in NumPy. I think we shouldn't PR this in at all until @arthurmccray has said his piece. Either way, I'll give my opinion and hopefully cover some things Arthur would probably have questions about.

1. I'm a little confused with why this is needed. I'm assuming downstream in your widgets you're copying `array` and making it into a `torch.Tensor` right? Why is there VRAM --> RAM --> VRAM transferring over and over again?  I think I need to understand what this memory constraint you're running into, because to me it's a little unclear why this is needed. If this can be clarified in why this bottleneck is happening that would be great. 
2. This is more a design choice, if the `array` is the only thing that needs to be operated in, I don't see why we couldn't just make a quick `def to(self, device)` function in `Dataset4DSTEM`. This bypasses making another `Dataset4dstem` class.

If we're really in a rush to implement this, we should probably go back and forth with this a little bit more, because I think making another `Dataset4DSTEM` specifically just for `torch` is not the answer.

src/quantem/core/datastructures/dataset4dstem_torch.py
cedriclim1: Can you check saving still works. I know Colin implemented some stuff that allows for `.zarr` saving with `torch` stuff.

bobleesj:
> I'm a little confused with why this is needed.

Regarding `VRAM --> RAM --> VRAM`, mostly for real-time microscope processing.

1. VRAM: I load 4DSTEM dataset to cupy (mallard/buffle), save about 10-20s of wall time on the scope.
2. RAM: I want to containerize `4DSTEM` dataset. I use `dataset4dstem`. Here, I must convert `cupy` into `numpy` -  cost 5-20s of wall-time
3. VRAM:  I want to use ptycho/ssb/widget, etc. I have to load that to VRAM (torch) - cost 5-20s of wall-time

ballpark numbers

This behavior is explained here: https://github.com/electronmicroscopy/quantem/issues/222

`Dataset4dstemTorch` can be a working, bandage solution that does not require huge refactoring for the time being and a test case. It can be a learning opportunity for everyone with this new behavior.

bobleesj:
I forgot to mention - I use my own "Dataset4dstemGPU" dataclass in my prototype to circumcent the VRAM-RAM-VRAM issue.  Having a container for VRAM would be nice so that I can use native `quantem` APIs instead of creating my own.

cedriclim1:
> Dataset4dstemTorch can be a working, bandage solution that does not require huge refactoring for the time being and a test case. It can be a learning opportunity for everyone with this new behavior.

Thanks for clarifying. You are right, this is a bandaid solution, but we probably should not also make this into a habit where everyone just starts initializing their datasets straight to `Dataset4dstemTorch`. 

I still think this shouldn't be in `core/datastructures`, and IMO any changes or additions there should be talked thoroughly with the core designers (Arthur/George), of which I am not. Arthur will be around next week (I think), so if this is not urgent it should be put off until he's back. 

I'll propose something else which might be OK for this to be implemented now. What do you think about following the `dataset_models` route instead see [here](https://github.com/electronmicroscopy/quantem/blob/dev/src/quantem/diffractive_imaging/dataset_models.py). In `widgets` you can just have some file `widgets/dataset_models.py` which has `Dataset4DSTEMWidgets` or something which has all the `torch` stuff you need, you can keep the current implementation and can also include `def from_dataset4dstem` for loading data, and a `def save_dataset4dstem` for saving back to a `Dataset4Dstem` or `Dataset4dstemwidgets` whatever. Would this way still cost too much I/O speed? I think this solves your issue potentially:

1. Load 4DSTEM dataset into `Dataset4dstemWidgets` straight to `torch`. 
2. If you want to do Ptycho, we might have to change some parts of the `dataset_models.py` to accept `Dataset4dstemWidgets`, but this would, _hopefully_, just be a straight `dataset.to(device)` kind of thing. 

I think this solution is good for now. If you need additional features that is specifically just for widgets, you could just implement here without overbloating `Dataset4dstem` or tainting `core/datastructres`. Moreover, if I start to add stuff to Tomography widgets, that are specifically only for the widget, this would be easy to do in this framework. 

> I forgot to mention - I use my own "Dataset4dstemGPU" dataclass in my prototype to circumcent the VRAM-RAM-VRAM issue. Having a container for VRAM would be nice so that I can use native quantem APIs instead of creating my own.

Yea, again can see why you're doing this and we should really think about integrating instant `torch` compatibility to the `core/datastructures` (I think this might be a you and me problem at some point, since Arthur and the others have other things to worry about). 


**Tl;dr**: Thoughts on going the dataset models way? Would be easy to integrate any widget-only functionality without tainting `core/datastructures`, and transferring to the core datastructures could be integrated into these dataset models. Also more extensible to other stuff: PDF, tomography, strain, etc...

cedriclim1:
Actually, we can probably do a `widgets/dataset_models` directory which has all the dataset widget models.

cedriclim1:
5/15/2026 - From our conversation please, propose a solution to making `Dataset4dstem`natively support `torch`!

bobleesj:
@cedriclim1 thanks Cedric and @henrygbell  for fruitful discussion during a sit-down

Yes, instead of creating of this short-term solution, we agree to see how we can both integrate `VRAM` and `RAM` in `Dataset4dstem` itself without creating another class (the original proposed solution https://github.com/electronmicroscopy/quantem/issues/222 - we are aligned on this direction. 

I will close this PR and create another PR with a proper long-term solution and while maintaining backward compatibility

and invite @arthurmccray and @gvarnavi for thorough review.

arthurmccray:
Looks like you guys ended up on the same solution I did--it's probably time to move `Dataset` to being torch-native. I'll leave a comment on #222 with more of my thoughts :)

bobleesj:
@arthurmccray thanks Arthur for your feedback and comment. Will do read your comments carefully and see how we can support torch/numpy both as simple as possible.

PR #230, opened 2026-05-22
`Show3D`, `Show3DSlices` PR for alpha test and user feedback

bobleesj:

DRAFT

`Show3D` - multiple

use-case: multislice ptycho recon trial comparison, 3D denoising performance

sepen_gif_20260521_233458

`Show3DSlices` - synthetic data for now. Real-data look too good. 

sepen_gif_20260522_002418

DRAFT

draft notebooks:

[show3d_simple.ipynb](https://github.com/user-attachments/files/28139034/show3d_simple.ipynb)
[show3dslices_synthetic.ipynb](https://github.com/user-attachments/files/28138930/show3dslices_synthetic.ipynb)


bobleesj:
@ceobrero I will add a few more clean-up commits for your test.

bobleesj:
@dasol-yoon -  for the ptycho data I shared, you can `gh pr checkout` this one and run the notebook I've shared. 

Please feel free to make a PR to `https://github.com/bobleesj/quantem/tree/widget-show3d` or provide any design feedback here.

bobleesj:
@henryhng  Looking for your feedback as well, at the code level if you find any inefficiencies. I will need to clean the code a bit but FPS on real-data has been my priority - espeically comapring trials across multislice ptycho recons.

bobleesj:
@cedriclim1 I initially had `Show3DVolume` but turns out ptycho multislice needed a bit different panel structure (like z-axis stretch), but thanks for your initial input and inspiration 

`Show3DVolume` is still available ([bobleesj.widget](https://github.com/bobleesj/quantem.widget/blob/main/src/quantem/widget/show3dvolume.py)and will make a separate PR after this one merged... or maybe a better idea is that you can just copy this and customize/build your own (maybe this is better?)

cedriclim1:
Sure - I can start playing around with your widget stuff. Are you thinking since this is for Ptycho we should branch it off to `Show3DTomo` and `Show3DPtycho` or something?

bobleesj:
Feedback from @cophus during GM

- [x] be able to remove plane in the 3D Volume

bobleesj:
> Sure - I can start playing around with your widget stuff.

@cedriclim1 As we discussed in the elevator, `Show3DVolume` is probably more general name for those who work with 3D objects, and `Show3DSlices` for 3D object that are "thin".

henryhng:
> @henryhng Looking for your feedback as well, at the code level if you find any inefficiencies. I will need to clean the code a bit but FPS on real-data has been my priority - espeically comapring trials across multislice ptycho recons.

@bobleesj I've noticed that when using the manual frame slider, js/colormaps.ts destroys and recreates all the GPU buffers on every frame change, so dragging the slider re-allocates per panel each step, the params/read buffers are persistent across `applySlots`, but `uploadData` itself recreates them

Similarly, dragging the slider re-sends a whole frame over the message channel each step which could be slow for big frames. I think a possible fix for this could be to use the same implementation as playback, where Python pre-ships a batch of frames into a buffer that lives in JS memory so no new message per frame is needed

bobleesj:
@henryhng Thanks for this. Pushing a new stack of commits addressing performance. Will re-visit your comments. Appreciate for the findings.

ceobrero:
@bobleesj Looking at the .html file for the denoised in-situ heating series, instead of having the FOV set to the initial frame dimensions, I think it would be better to use a larger FOV that includes padding around the entire frame, so that when scrubbing through the dataset, every single frame is in full view. Right now, several frames are cropped on the sides during alignment, resulting in information loss. 
image

Frames should look something like this, where the padding is larger than the bounds of the aligned frames: 
image

After making a line profile, when I change the zoom, the line profile doesn't scale with the zoom, offsets, and clips into another quadrant:
image image image
Also, I can't seem to find the ROI button...?

This might be personal preference, but is there a way to resize the widget so that it automatically fits into my browser dimensions? It seems set to a specific size and currently, when I click FFT it cuts offscreen, and I have to make my browser Zoom to see the color bar (my normal 100% Zoom vs. 50% Zoom):
image
image

For exporting, would it be possible to have the option to also export with a scale bar? Also, exporting PNGs to a zip file and exporting a GIF either takes a really long time (progress bar?) or doesn't work, leaving "..." instead or "export".
image

bobleesj:
@ceobrero Thanks for testing

> FOV that includes padding around the entire frame

- [x] Regarding this, I will see if it is easy to add "padding" parameter. `Show3D` isn't meant to be used a prepreocessing but I think that is a reasonable use-case. 

> the line profile doesn't scale with the zoom, offsets, and clips

- [x] Will fix line profile. 

> Also, I can't seem to find the ROI button...?

- [x] Will remove ROI for Show3D with multi-panel - will be too clutterered. 

> resize the widget so that it automatically fits into my browser dimensions? 

[next PR?] a new feature that may break existing behavior - so next PR 

> either takes a really long time

[next PR] Will remove export - again multi panel Show3D may break existing

> Export with a scale bar? 

[next PR] will add scale bar in the next PR

bobleesj:
I will make a few more commits in the next 1-2 and I will test and share another `.html ` for your review @ceobrero

ceobrero:
@bobleesj Frames still get cropped when scrubbing through the data stack:
image

The x and y cursor coordinates seem to be flipped. Moving up and down changes the first index while moving side to side changes the second index. 
image
The second index is also currently treated as if the four quadrants were horizontally adjacent, like one long image (800 is unrealistic, given the frame size is ~237 px):
image

Turning contrast off creates four separate intensity histograms that clip into the slider:
imageimage

bobleesj:
> @bobleesj Frames still get cropped when scrubbing through the data stack:

This is data itself. Unless you want to crop the data itself. If that's the case, we can add a function within `Show3D` . Pls lmk if this can be useful.

> The x and y cursor coordinates seem to be flipped. 

Going with row/col convention for STEM images. So single source of truth based on geometry.

> like one long image (800 is unrealistic, given the frame size is ~237 px):

- [x] Will fix. Each panel should be independently

> Turning contrast off creates four separate intensity histograms that clip into the slider

- [x] Yeah, I will just make the width bigger so that they don't overlap. 

Thanks!

bobleesj:
May 29 2026 `Show3D` feedback from @cophus 

- [x] make it HTML exportable for Show3D
- [x] make it GIF exportable for Show3D

henrygbell:
This is slightly outdated but here are a few comments I had for improvement:
- [x]  One feature that would be neat is to add the option to display the average of the images in a moving window. This is really helpful if you have noisy data. Basically this is almost like binning in time. I think digital micrograph does this pretty well when you view videos. Maybe you can add a slider near the 'fps' slider where you can choose the size of the moving window in time.
- [x] It would be incredibly useful if the profile feature created a 2D image to the side (like how FFT pops a window up) where one dimension is the distance along the line, and the other is the time axis. 
Small thing: the bar in the bottom to control which image to see changes sizes when the numbers to the right of it have more or less digits. I think you can fix it easily by fixing the size of the displayed numbers.
- [x] It seems like the Show3D scale bar did no inherit my units/sampling for the dataset3D class, but potentially I did it wrong. Here is my code for that:

```
datasetQ =  Dataset3d.from_array(
    array = dataset_binned,
    name = "UEM scan",
    sampling = [15, pixel_size_micron, pixel_size_micron],
    units = ["ps", "micron", "micron"],
)
Show3D(datasetQ)
```

bobleesj:
@henrygbell Thanks Henry. Most are done. Will keep you posted.

bobleesj:
@henryhng Thanks for feedback. All fixed and kymogram added. Super useful for iterative ptycho depth res.

Screenshot 2026-06-01 at 9 40 48 AM

bobleesj:
20260601 feedback from @cophus 

- [x] Replace "Row" and "Col" plane and projection with a single plane with arbitrary angle about the z-axis, and control the angle via sliders.

bobleesj:
@ceobrero @henrygbell thanks everyone for great feedback. to provide better dev experience and given we have 5-6 widget devs, we will migrate our code over to `electronmicroscopy/quantem.widget`.

PR #239, opened 2026-06-01
Fix latent bugs in tomography optimizer wiring, constraints, and radon

cedriclim1:
# What problem this PR addreseses

step_optimizers and zero_grad_all looped over optimizer_params and stepped both the object and pose optimizers on every pass, so with pose optimization enabled each optimizer took two Adam steps per batch. Gate by key, matching step_schedulers.

Also:
- pass num_iter to set_schedulers on the reset_dset path so cosine/linear/ exponential schedulers get a valid T_max
- scheduler_params setter no longer mutates the caller's dict
- drop the non-existent tv_plane key from ObjINRConstraints.soft_constraint_keys (it crashed Constraints.__str__)
- ObjectPixelated.get_tv_loss now takes TV over the trailing spatial dims, so it handles a 3D volume, obj_view's [1, D, H, W], and a multimodal [C, D, H, W]
- remove an unreachable duplicate branch in TomographyINRDataset.forward
- align iradon_torch's default theta with radon_torch / scikit-image (endpoint excluded)

Add regression tests covering each fix.

# What Reviewer Should Do

Nothing - all tests pass checking for these fixes.

### DIFF
--- a/src/quantem/tomography/dataset_models.py
+++ b/src/quantem/tomography/dataset_models.py
@@ -486,8 +486,6 @@ class TomographyINRDataset(TomographyDatasetConstraints, Dataset):
             return shifts, torch.zeros_like(z1), torch.zeros_like(z3)
         elif self.learn_tilt_axis:
             return torch.zeros_like(shifts), z1, z3
-        elif self.learn_shift and self.learn_tilt_axis:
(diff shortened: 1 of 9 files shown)


arthurmccray (approved):
lgtm

PR #208, opened 2026-04-07
drift: speed up `align_nonrigid()` to handle 2048×2048, 4096x4096 pairs in seconds

bobleesj:
### What this PR does

No API change, no infrastructure change. 

Subsequent PR for optimizing `non-rigid` drift correction after https://github.com/electronmicroscopy/quantem/pull/206. Back then, I did implement torch for `non-rigid`  in https://github.com/electronmicroscopy/quantem/pull/133 with 2-3x but here we are looking at up to 400x shown below:

Performance:

benchmark_nonrigid_bars

benchmark_nonrigid_full_results
                                                                                                                                                                     
  | Size | SciPy time | Adam time |  LBFGS time | Adam speedup | LBFGS speedup |
  |------|-----------:|----------:|-----------:|-------------:|--------------:|
  | 128×128   |   3.33s | **0.097s** |   0.203s |  **34×** |  16× |
  | 256×256   |   5.62s | **0.096s** |   0.187s |  **59×** |  30× |
  | 512×512   |  10.26s | **0.101s** |   0.147s | **102×** |  70× | 
  | 1024×1024 |  22.18s |   0.078s | **0.177s** | **284×** | 125× |
  | 2048×2048 |  62.02s |   0.244s | **0.142s** |    254×  | **437×** |

### What should reviewer do

This PR will be open for review available after https://github.com/electronmicroscopy/quantem/pull/206 merged. 

I will share another layer of comment TBA. 

For now, I want to make a PR so that I don't write too much in each branch and reduce brain burden.

PR #206, opened 2026-04-06
drift: speed up `align_affine()` to handle 2048×2048, 4096x4096 pairs in seconds

bobleesj:
### What this PR does

No API change, no architecture change from the initial knot/warping design by @cophus

Here we gain two orders of magnitude faster on GPU so that can we can drift correct on the scope in seconds after acquisition. For example, 2k image used to take ~10 mins, now, it can be done in 1-3 seconds on a GPU.

Performance:

Screenshot 2026-04-06 at 10 50 15 AM

  Image size     | Before | After  | Speedup                                                       
  ---------------|----------------|-----------------|--------                                                       
  256×256        |       3.0s     |      0.03s      |  100×                                                         
  512×512        |      27.0s     |      0.10s      |  270×                                                         
  1024×1024      |     109.0s     |      0.40s      |  273×                                                         
  2048×2048      |     436.0s     |      1.43s      |  305×    

The speedup is gained primarily by batching operations on a set of candidate drift vectors.

### What should reviewer do

I wrote parity tests for `drift_utils.py` against `scipy` and `numpy`.  I also tested against my own data I collected with @cophus  and Amy. I don't suggest spending too much time reviewing (~10-20 mins if you have time) since I need to make a few PRs on `non-rigid` and memory management to `upstream/drift`.

But please feel free to run your existing notebook to ensure you also get the same result.

Also tagging @henrygbell for  MAPED implementation since you were on the business of speeding things as well.

PR #248, opened 2026-06-11
Torch-native PDF pipeline + origin finder improvements

ehrhardtkm:

Follow-on to #177 (the polar transform + PDF pipeline that merged into `nanobeam`). This PR brings that work onto `dev` and lands the changes I've made since.

**Conceptually unchanged** — same `Polar4dstem` datastructure, same `PairDistributionFunction` shape, same overall workflow (origin → polar → PDF). The public API should be the same, except for a few minor changes.

**Changes since #177:**

- **Fully torch-native, dual-backend pipeline.** The whole pipeline accepts numpy or torch input on the existing `Dataset`, device is inferred from the input, and there are no GPU↔CPU round-trips inside the pipeline. numpy vs tensor input give identical results.

- **Background fitting.** Rewritten with an analytic Jacobian that is noticeably faster than the previous implementation. Added a batched `fit_bg_batched` path for full-dataset use, which the downstream PDF widget (PR3) needs to recompute backgrounds interactively. Also includes some small fixes I picked up while running this on real data.

- **Origin finding.** The existing finder's hardcoded margin requirement caused it to error out on smaller detectors, so I made it the margin dynamic and added a 2D-quadratic subpixel refinement (renamed to `find_origin_angular_grid`). I also added a secondary option, `find_origin_angular_descent`,  a margin-free, COM-anchored descent finder. It's slower but I've found it slightly more reliable on noisier data. Both are dispatched from `Polar4dstem.from_data(origin_method="grid"|"descent")`.

- **Tests** in `tests/diffraction/test_polar.py` are parametrized over both origin finders and add coverage for the new code paths.

**Out of scope (follow-on PRs to shortly follow):**
- Elliptical-distortion fit + reciprocal-pixel calibration → PR2, stacks on this branch.
- ShowPDF widget vendor → PR3.

Merge after review. Reviewing commit-by-commit is likely easier than the unified diff — each commit is a single conceptual change.
    - [ ] Documentation (e.g., tutorials, examples, README) has been updated.
    - [ ] A tracking issue or plan to update documentation exists.
- [ ] This PR affects internal functionality only (no user-facing change).

### DIFF
--- a/src/quantem/core/datastructures/__init__.py
+++ b/src/quantem/core/datastructures/__init__.py
@@ -2,6 +2,7 @@ from quantem.core.datastructures.dataset import Dataset as Dataset
 from quantem.core.datastructures.vector import Vector as Vector
 
 from quantem.core.datastructures.dataset4dstem import Dataset4dstem as Dataset4dstem
+from quantem.core.datastructures.polar4dstem import Polar4dstem as Polar4dstem
 from quantem.core.datastructures.dataset4d import Dataset4d as Dataset4d
 from quantem.core.datastructures.dataset3d import Dataset3d as Dataset3d
 from quantem.core.datastructures.dataset2d import Dataset2d as Dataset2d
--- a/src/quantem/core/datastructures/polar4dstem.py
+++ b/src/quantem/core/datastructures/polar4dstem.py
@@ -1,0 +1,139 @@
+from typing import Self
+
+import numpy as np
+import torch
+from numpy.typing import NDArray
+
+from quantem.core.datastructures.dataset4d import Dataset4d
+
+
+class Polar4dstem(Dataset4d):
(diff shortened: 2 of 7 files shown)


bobleesj:
@ehrhardtkm before I review, it seems like being merged to `upstream/dev` instead of `upstream/nanobeam`. Has `upstream/nanobeam` containing pdf code merged to `upstream/dev`?

bobleesj:
Another question I had was from the PR comment was - what's approx speed/improvement given various sizes? Mostly likely it will be either 512x512x192x192 or 256x256x192x192 to be practical, just ballpark numbers. 

Ref: 

- https://github.com/electronmicroscopy/quantem/pull/206
- https://github.com/electronmicroscopy/quantem/pull/208

PR #249, opened 2026-06-13
Direct ptycho visualize and rotation

smribet:

This PR makes direct and iterative ptycho more consistent by changing direct ptycho to degrees and adding a visualize function. 

FWIW I did some testing yesterday, and I think the direction of aberrations is consistent between abTEM, direct, and iterative ptycho, but others should confirm as well. 

Merge the code and provide feedback.
visualize_output

### DIFF
--- a/src/quantem/diffractive_imaging/direct_ptycho_utils.py
+++ b/src/quantem/diffractive_imaging/direct_ptycho_utils.py
@@ -495,7 +495,10 @@ def fit_aberrations_from_shifts(
     gpts: tuple[int, int],
     sampling: tuple[float, float],
 ) -> dict[str, float]:
-    """ """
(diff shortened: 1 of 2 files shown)


gvarnavi (approved):
Thanks Steph, looks good to me!

Note aberration axes still need to be specified in radians, but I think we should leave this to match abTEM.

```python
# ground truth
known_aberration_coefs = {
    "C10": 200,
    "C12": 100,
    "phi12": np.deg2rad(11)
}
known_rotation_angle = -13
```

Updated tutorials to match: https://github.com/electronmicroscopy/quantem-tutorials/pull/11

Feel free to merge both when ready.

smribet:
I agree aberrations in radians makes sense. Thanks for the feedback @gvarnavi!

PR #256, opened 2026-07-01
Maped torch implementation

henrygbell:

The existing MAPED class (https://github.com/electronmicroscopy/quantem/pull/169) runs on NumPy/CPU, which makes the alignment + merge pipeline slow on the large 4D-STEM tilt/precession stacks it's meant for. This PR adds MAPEDTorch, a PyTorch port of the full MAPED workflow that runs on GPU (or CPU) and operates directly on torch.Tensor 4D-STEM volumes.

[Describe the problem this PR solves. Focus on inputs/outputs.
Attach screenshots, plots, and before/after comparisons.]

[Show the function signatures, class interfaces, and how a scientist
would use this in a notebook. The API is what a scientist actually
types. Get this right first.]

This PR implements the following methods:

1. preprocess()
    - Computes bright field and mean diffraction patterns for all datasets
2. diffraction_origin()
    - Computes the origin of the mean diffraction patterns
3. diffraction_align()
    - Uses cross-correlation based registration to align the mean diffraction patterns and saves the shifts
4. real_space_align()
    - Uses cross-correlation based registration to align the bright field images and saves the shifts
5. merge_datasets()
    - Apply the shifts computed in the last step and merge the datasets. Returns a dataset_4DSTEM object with the merged data

API — how a scientist uses it in a notebook:
```
from quantem.diffraction import MAPEDTorch

# datasets: list of torch.Tensor, each shape (R, C, H, W), all same dtype + device
maped = MAPEDTorch.from_datasets(datasets)

maped.preprocess(plot_summary=True)                      # -> mean BF
maped.diffraction_origin(sigma=None, plot_origins=True)  # find DP origins
maped.dscan_align(iterations=5)                          # optional descan correction

maped.diffraction_align(num_iter=30, edge_blend=2.0, upsample_factor=100)
maped.real_space_align(num_iter=30, edge_blend=5, hanning_filter=True,
                       pad_val="median", shift_method="bilinear")

merged_ds = maped.merge_datasets(shift_method="bilinear", batch_size=4)  # -> Dataset4dstem
```

Attached is a validation notebook which executes the whole pipeline against ground truth and is the easiest way to review correctness:  [maped_torch_validation.ipynb](https://github.com/user-attachments/files/29561645/maped_torch_validation.ipynb)

Validation notebook workflow:
1. It builds N synthetic datasets with known real-space and diffraction-space shifts applied per dataset.
2. It runs the pipeline and checks that MAPEDTorch recovers those shifts. The residual center(applied + recovered) should be ≈ 0 for every dataset; the notebook asserts a max residual < 0.2 px for both diffraction_shifts and real_space_shifts.
3. It merges the aligned stack and confirms (via shift-invariant max NCC) that the merged result recovers the underlying structure.

Current results of this PR's MAPED code

```
# Simulated dataset parameters
N = 5               # number of datasets
NOISE = 0.05        # fraction of peak amplitude added as gaussian noise 
shift_tol = 0.2      # tolerance in pixels for shift residual
ncc_tol = 0.95      # tolerance for merged dataset to ground truth ncc
```

image

```
# Validation results
Diffraction alignment : PASS (residual 0.0715 px)
Real-space alignment  : PASS (residual 0.1012 px)
Merge structure recov.: PASS (NCC 0.9989)

ALL CHECKS PASSED 
```

Please use this notebook for future updates to the MAPED pipeline to assess correctness.

Please look at:
- The validation notebook above
- GPU device handling

If possible:
- Test on your own data

### DIFF
--- a/src/quantem/diffraction/maped.py
+++ b/src/quantem/diffraction/maped.py
@@ -277,7 +277,7 @@ class MAPED(AutoSerialize):
             shift_rc, G_shift = weighted_cross_correlation_shift(
                 im_ref=G_ref,
                 im=G,
-                weight_real=im_weight * 0.0 + 1.0,
+                weight_real= 1.0,
                 upsample_factor=int(upsample_factor),
                 fft_input=True,
                 fft_output=True,
@@ -1166,19 +1166,32 @@ class MAPEDTorch(AutoSerialize):
 
     def diffraction_align(
         self,
+        num_iter: int = 5,
         edge_blend: float = 16.0,
         padding=None,
         pad_val: str | float = "min",
         upsample_factor: int = 100,
         weight_scale: float = 1 / 8,
+        max_shift=None,
         plot_aligned: bool = True,
         **plot_kwargs: Any,
     ) -> MAPED:
         """
-        Align mean diffraction patterns using weighted cross-correlation in Fourier space.
+        Align mean diffraction patterns using iterative average-reference
+        cross-correlation in Fourier space.
+
+        Each iteration phase-shifts every diffraction pattern by the current shift
+        estimate, forms a common average reference from all patterns, then
+        re-correlates each pattern against that reference and accumulates the
+        residual shift. This mirrors ``real_space_align`` so that after the first
+        iteration only small residual shifts are measured, where subpixel
+        cross-correlation is most accurate. A single pass (num_iter=1) reproduces
+        the old behavior of measuring each (large) shift once against a mean.
 
         Parameters
         ----------
+        num_iter : int
+            Number of refinement iterations.
         edge_blend : float
             Tukey window edge taper (pixels).
         padding : int or tuple, optional
@@ -1188,7 +1201,11 @@ class MAPEDTorch(AutoSerialize):
         upsample_factor : int
             Subpixel upsampling factor for correlation peak estimation.
         weight_scale : float
-            Radial weight falloff scale (fraction of mean DP size).
+            Radial weight falloff scale (fraction of mean DP size). Currently unused.
+        max_shift : float, optional
+            Largest expected shift magnitude (pixels), used to size the zero-padded
+            correlation canvas so large shifts do not wrap around. If None, it is
+            estimated from the spread of ``diffraction_origins``.
         plot_aligned : bool
             If True, plot aligned mean diffraction patterns.
         **plot_kwargs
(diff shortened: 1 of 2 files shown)


bobleesj:
@henrygbell beautiful github workflow. clean simple, thanks for the tests. I will review later this week!

gvarnavi:
Quick comment on the class naming: Presumably the ambition is to make most (all) quantem functionality torch-first, so I would avoid class names like `MAPEDTorch`. If you want to keep the existing numpy class functionality, then I would rename that one something like `MAPEDNumpy` instead.

src/quantem/diffraction/maped.py
```diff
@@ -890,3 +895,1801 @@ def shift_images(
...
+                shifts_rc=self.diffraction_shifts,
+                edge_blend=float(edge_blend),
+                padding=padding,
+                pad_val=pad_val,
+            )
+            show_2d(im_aligned.unbind(0), **plot_kwargs)
+
+        return self
+
+    def real_space_align(
```
bobleesj, line 1357: we need a good documentation on which parameters (and magnitude) to use. are there any chances where these parameters matter? these human lessons should be documented in general. 

p.s. the goal is to not using/writing any parameters when possible - all automated if possible. but that's not really possible in practice.

bobleesj: generally for all other function as well - scientists wll run the pubic API in jupyter notebook, they find their results are bad, then they will revert to this API doc, 

we want this API doc to provide clues/lessons for them. - smae goes for LLM/AI.

src/quantem/diffraction/maped.py
```diff
@@ -890,3 +895,1801 @@ def shift_images(
     out[den == 0.0] = 0.0
 
     return out
+
+
+class MAPEDTorch(AutoSerialize):
```
bobleesj: as @gvarnavi mentioned - MAPED, don't think we have to call it `Maped`, etc. MAPED

src/quantem/diffraction/maped.py
```diff
@@ -890,3 +895,1801 @@ def shift_images(
...
+                    )
+
+                    wi_sample = torch.nn.functional.grid_sample(
+                        w_rs_reshaped,
+                        grid_batch,
+                        mode="bilinear",
+                        padding_mode="zeros",
+                        align_corners=True,
+                    )
+
```
bobleesj: try not to have any extra empty lice spaces - we want code to read like paragraphs, each paragraph, each thesis/reason to exist.

src/quantem/diffraction/maped.py
```diff
@@ -890,3 +895,1801 @@ def shift_images(
...
+    if imageCorr.ndim != 3 or xyShift.ndim != 2:
+        raise ValueError("imageCorr must have shape (M, N) or (B, M, N), and xyShift must match")
+    if imageCorr.shape[0] != xyShift.shape[0]:
+        raise ValueError("imageCorr and xyShift batch dimensions must match")
+
+    device = imageCorr.device
+    _, M, N = imageCorr.shape
+    pixelRadius = 1.5
+    numRow = int(math.ceil(pixelRadius * upsampleFactor))
```
bobleesj, line 2401: numRow, row_coords, etc. keep naming consistent throughout

src/quantem/diffraction/maped.py
```diff
@@ -890,3 +895,1801 @@ def shift_images(
...
+                vr_p = dp_b[batch_idx, row_safe + 1, col_peak].clamp(min=1e-6).log()
+                denom_r = vr_m + vr_p - 2.0 * vr_0
+                dr = torch.where(
+                    (denom_r < -1e-6) & (row_peak > 0) & (row_peak < H_dp - 1),
+                    ((vr_m - vr_p) / (2.0 * denom_r)).clamp(-1.0, 1.0),
+                    torch.zeros(B, device=device),
+                )
+
+                # log-parabolic sub-pixel refinement — col direction
+                col_safe = col_peak.clamp(1, W_dp - 2)
```
bobleesj, line 2615: yup, as  mentioned, this is a good code block where no extra line spaces, but only add one when there is a reason to add.

bobleesj (commented):
@henrygbell left comments for strengthening API doc to guide humans and some minor formatting

henrygbell:
@bobleesj Thanks for your comments. They have been addressed in the recent commits, I also updated the validation notebook for the MAPEDTorch -> MAPED change:

[maped_torch_validation.ipynb](https://github.com/user-attachments/files/29680533/maped_torch_validation.ipynb)

bobleesj:
@henrygbell merging. I've tested the code a few times. I will make another PR cleaning up NumPy, etc and adding some unit tests.

PR #269, opened 2026-07-23
Diffractive imaging

arthurmccray:

Placeholder for pre-M&M pull from diffractive_imaging into dev.


arthurmccray:
@smribet wanna review this? it's mostly changes we've been using for a while, so I'm not too worried about it. I'm sure there'll be one or two bugs that we find and need to fix, but i'd like to get dev up to date for M&M.

smribet (approved):

PR #271, opened 2026-07-30
Fixes for strain mapping

cophus:
Some bug fixes for disk detection strain mapping.


cophus:
Superseeded by another PR 274

PR #272, opened 2026-07-31
Vector updates

arthurmccray:
Move Vector to torch, add to_polars()

Vector is only used by a few folks, so it seemed easiest to just break things by moving to torch rather than doing a slow migration. 

Breaking changes

- .array → .tensor. The 0D cell accessor returns a torch.Tensor. Note this is one cell, unlike Dataset.tensor which is the whole payload.
- flatten() and all arithmetic return tensors. NumPy arrays, lists and scalars are still accepted as input everywhere; .numpy() converts back (read-only, matching Dataset.numpy()).
- Default dtype is torch.float32 (was float64). Pass dtype=torch.float64 to from_shape/from_data if you need double.
- NumPy ufuncs are disabled. np.sin(v) raises TypeError; use torch.sin(v). __torch_function__ replaces __array_ufunc__, gated on an explicit pointwise-op allowlist — so torch.sin(v) returns a Vector while torch.sum(v) and torch.t(v) return plain tensors. Reductions now work, which they didn't under the ufunc protocol.

New

- to_polars(dim_names=None) — flattens to a DataFrame with the fixed-grid location as leading dim_0…dim_n integer columns. Honors the current view (both [] and select_fields), and reports root-grid coordinates so a row always says which scan position it came from. polars is an optional dep (quantem[dataframe]), lazily imported, not in dev/test.
- .device, .to(device), .numpy(), dtype=/device= on the factories. Offset bookkeeping stays on CPU by design — it's read one scalar at a time, so keeping it on-device would sync per cell access.

Bug fix

select_fields("c","b","a") returned data in storage order while .fields reported the requested order. Pre-existing (same logic in the NumPy version); fixed and tested.

Serialization

save() writes the row buffer as compressed Zarr rather than an uncompressed torch.save blob, and a new _post_load hook on AutoSerialize rehydrates to CPU tensors. Consequences: pre-migration files still load (keeping their original dtype), and a GPU-saved Vector loads onto CPU.

Known limitation: this only applies when the Vector is the save root. AutoSerialize walks nested objects via _recursive_save, so a Vector held as an attribute of another serializable object still round-trips correctly but writes uncompressed (~48% larger). The real fix is teaching _serialize_value to store plain non-grad tensors as Zarr arrays — which would also fix tensor-backed Dataset, currently saving fully uncompressed. Documented in save(), deliberately out of scope here.

Make sure it works for your workflows, make fixes. 
Take a look at the tutorial: https://github.com/electronmicroscopy/quantem-tutorials/blob/main/tutorials/core/vector.ipynb 

    - [x] Documentation (e.g., tutorials, examples, README) has been updated.


arthurmccray:
@cophus just pinging you as discussed so we can get this into dev :)

PR #273, opened 2026-07-31
Fitting models clean

maclariz:

This creates a new set of functions in diffraction to provide Digital Dark Field imaging, both with manual selection and cluster analysis.  It is all in one py file split into three parts:
1) import of data into Vector class
2) regular DDF imaging that reproduces some functions in py4DSTEM
3) the starts of cluster analysis (mainly L1 clustering)

All can be built out once the module has been accepted

The initial import is based on my own centre finding, which I can provide, but you may have code elsewhere in the repo that produces the desired kx and ky centre positions, in which case just adjust my function to use that result.


maclariz:
And don't change "vector.py" in response to this.  I initially changed it and then changed it back, so it looks changed but isn't.  All the new stuff in in digital_dark_field_cluster.py and __init__.py was changed in diffraction to import the new functions into the repo

maclariz:
So, functions rebuilt to work with a new image reduce method in Vector and all built out to L2 clustering with two methods.  The image reduce could be generalised, and an option for reducing to a diffraction pattern could also be made.

src/quantem/diffraction/__init__.py
```diff
@@ -1,4 +1,5 @@
 from quantem.diffraction.bragg_vectors import BraggVectors as BraggVectors
 from quantem.diffraction.strain import StrainMap as StrainMap
 from quantem.diffraction.strain_autocorrelation import StrainMapAutocorrelation as StrainMapAutocorrelation
-from quantem.diffraction.model_fitting import ModelDiffraction as ModelDiffraction
\ No newline at end of file
+from quantem.diffraction.model_fitting import ModelDiffraction as ModelDiffraction
+from quantem.diffraction.digital_dark_field_cluster import *
```
arthurmccray, line 5: Please do not `import *` , if you want certain features elevated in the namespace please import specific items (as is done in this file)

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
+import numpy as np
+
+import matplotlib.pyplot as plt
+from matplotlib.colors import LinearSegmentedColormap, Normalize, PowerNorm
+import matplotlib.gridspec as GridSpec
+import matplotlib.patheffects as path_effects
+
+from tqdm import tqdm
+
+from sklearn.cluster import DBSCAN
```
arthurmccray, line 10: `sklearn` isn't a dependency of `quantem`, so please guard the import with an error message saying something like "if you want to use this feature, please install `scikit-learn` to your environment"

arthurmccray, line 10: `sklearn` isn't a dependency of `quantem`, so please guard the import with an error message saying something like "if you want to use this feature, please install `scikit-learn` to your environment"

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+    Returns
+    -------
+    """
+    
+    if figax is None:
+        fig, ax = plt.subplots(1, 1, figsize=(6, 6))
+    else:
+        fig, ax = figax
+        assert isinstance(fig, Figure)
+        assert isinstance(ax, Axes)
```
arthurmccray, line 442: these lines throw an import error

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+    assert level in [1,2], "level must be either 1 or 2"
+    level_label = f"L{level}labels"
+
+    labels = pointsvector.select_fields(level_label).flatten().astype(int)
+    unique_labels, all_cluster_sizes = np.unique(labels, return_counts=True)
+    shape = pointsvector.shape
+    if ordering == "sequential":
+        cluster_list = unique_labels[1:]
+    elif ordering == "size":
+        cluster_list = L1_unique_labels[1:][np.argsort(L1_all_cluster_sizes[1:])[::-1]]
```
arthurmccray, line 529: gives a `NameError`

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+        The raw Vector that was run through clustering 
+    maskstack: np.ndarray
+        A single mask or stack of masks selecting one or more clusters
+    Returns
+    -------
+    maskstack: no.ndarray
+
+    """
+    assert isinstance(maskstack, np.ndarray), "the maskstack must be a numpy array"
+    assert maskstack.shape[-1] == pointsvector.flatten.shape[1], "the mask size does not match the Vector size"
```
arthurmccray, line 613: should be `.flatten()`, and i'm not actually sure it's comparing row to row

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+    """
+    assert "L1labels" in pointsvector.fields, "This Vector does not appear to have been clustered"
+
+    rxy = pointsvector.select_fields("rx","ry").flatten()
+    I = pointsvector.select_fields("intensity").flatten()
+    L1labels = pointsvector.select_fields("L1labels").flatten().astype(int)
+
+    L1_unique_labels = np.unique(L1labels)[1:]
+
+    COMs = np.zeros_like(np.vstack((L1_unique_labels,L1_unique_labels)).T)
```
arthurmccray, line 644: are COMs supposed to be integer valued? this will truncate sub-pixel offsets

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+        A copy of the original Vector, with an additional field for L1labels.  It may be shorter than 
+        pointsvector if radial filtering has been applied
+    
+    '''
+    for item in fields:
+        assert item in ["rx", "ry", 'kx', "ky", "kr", "kphi"], "field not found in [rx, ry, kx, ky, kr, kphi]"
+    assert len(scaling)==len(fields), "the scalings and fields must have the same number of entries"
+
+    # We need to return a new Vector as it is changing length once we select only part of the data
+    pointsvector2 = pointsvector.copy()
```
arthurmccray, line 372: this line isn't used

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+    pointsarray = (np.array(scaling)*pointsvector2.select_fields(*fields).flatten())
+
+    db = DBSCAN(eps=eps, min_samples=min_samples)
+    db.fit(pointsarray)
+    pointsvector2.add_fields('L1labels',db.labels_)
+    if plot:
+        plot_L1_clusters_kspace( 
+            pointsvector2, 
+            fields, 
+            kr_max_plot=int(pointsvector2.select_fields('kx').flatten().max()*1.05)
```
arthurmccray, line 386: is this meant to be `kr` not `kx`?

src/quantem/diffraction/digital_dark_field_cluster.py
```diff
@@ -0,0 +1,769 @@
...
+        fields = ["kx","ky"]
+    pointsvector = Vector.from_shape(
+        shape=Rshape,
+        fields=("rx", "ry", "kx", "ky", "kr", "kphi", "intensity"),
+        units=("pixels", "pixels", "pixels", "pixels", "pixels", "degrees", "counts"),
+        name="diffraction_vectors",
+    )
+ 
+    for rx in tqdm(range(Rshape[0])):
+        for ry in range(Rshape[1]):
```
arthurmccray, line 51: I think there should be a way to vectorize this

arthurmccray (changes requested):
Hi Ian, overall looks good! I'm leaving a few comments in the files where errors would occur. A few more general things: 
- We are also in the process of moving `Vector` over to being base `torch`, (see https://github.com/electronmicroscopy/quantem/pull/272) That will cause a problem next time you try and merge from dev, but I think an LLM should be able to adapt what you've added to `torch` without many issues. I don't think this needs to be solved now. 
- a lot of the things i point out in comments would be caught with a simple pre-commit. I would suggest checking out the [CONTRIBUTING.md](https://github.com/electronmicroscopy/quantem/blob/dev/CONTRIBUTING.md) or just install and run `ruff`. 
- as a smaller thing, please refer to [PEP8 ](https://peps.python.org/pep-0008/#naming-conventions)for a style guide and conventions for naming (e.g. in the `DBSCAN_L2` function) 
- please write a few tests for DDF (or have an LLM do so for you)

src/quantem/diffraction/digital_dark_field_cluster.py
maclariz: I really suggest that you consider having scikit-learn as a dependency so we can leverage the tools they have built rather than having to reinvent the wheel ourselves.
Of course, that means we might have to take torch tensors and transform to numpy to access this, but my initial attempt at getting DBSCAN written for torch didn't work well.  Maybe this can be done eventually, but until then, scikit-learn is the best show on the road for clustering algorithms.

src/quantem/diffraction/digital_dark_field_cluster.py
maclariz: Yes, this is intentional.  It rather assumes the detector is square, which it probably is.  In the rare cases where it is not (e.g. Merlin T4 in frame-based mode), kx_max > ky_max, so this sets the maximum radius used in deciding maximum axis limits to the largest k value possible +5%

src/quantem/diffraction/__init__.py
maclariz: fair dos.  Will name everything explicitly.

src/quantem/diffraction/digital_dark_field_cluster.py
maclariz: easy to fix

PR #280, opened 2026-08-20
Preserve FP32 coordinates under BF16 in tomography and for INRs

ccasert:

This PR fixes BF16 instability in INRs by keeping rays and sampler coordinates (for kplanes) in FP32.

Coordinates are normalized to `[-1, 1]`. At high enough resolution, training INRs with BF16 can cause instability because the spacing between points can become less than BF16 can represent, so adjacent samples can collapse to the same value.

It also applies BF16 consistently during training, validation, and pretraining for tomography; previously reconstruction used BF16 only during validation.

Review the FP32 and BF16 coordinate behavior.

  - [x] This PR introduces a public-facing API change (`use_bfloat16`).

### DIFF
--- a/src/quantem/core/ml/models/kplanes.py
+++ b/src/quantem/core/ml/models/kplanes.py
@@ -309,6 +309,13 @@ class KPlanes(PPLR, TensorDecompositionModel):
 # ---------------------------------------------------------------------------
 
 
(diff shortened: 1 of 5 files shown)


cedriclim1 (approved):
LGTM!

PR #281, opened 2026-08-21
Direct ptychography: real-space montage, empirical probes, non-electron units

gvarnavi:
## Overview

Adds `DirectPtychographyMontage`, a second direct-ptychography class that accumulates the scan onto a real-space canvas instead of Fourier transforming over it, and extends both classes to handle probes and wavelengths that the electron-only code path could not express.

`DirectPtychography` gains an ungridded constructor, a test suite (it had none), and three bug fixes — one of which meant no reconstruction could be saved and reloaded.

## The idea

Every kernel in `DirectPtychography` is a multiplier on `G(k, q)`, the virtual bright-field stack Fourier transformed over the scan. That requires the scan to lie on a regular grid, and it requires the whole transformed stack in memory at once.

`DirectPtychographyMontage` drops the scan-space transform: each bright-field image is deposited onto one shared canvas at its own probe position. The detector axis is handled identically. This buys three things that are not available in the Fourier formulation:

- **Ungridded scans**, with no resampling step — positions are used as measured.
- **Position-dependent defocus**, for a tilted sample. A Fourier multiplier is global over the scan by construction and cannot express this.
- **Streaming memory.** The Fourier class materializes `(N_bf, Ry, Rx)` complex64 — 34 GB at 167k bright-field pixels on a 128×170 canvas, where the montage needs ~800 MB.

## Which class to use

| | |
|---|---|
| gridded scan | `DirectPtychography` — exact, and cheapest at one scan FFT |
| ungridded scan | either: `DirectPtychography.from_dataset3d` regrids onto a lattice first, the montage never grids |
| sub-pixel positions matter | montage — regridding discards them (0.997 against 0.982 CTF correlation over the extension band) |
| scan both masked and upsampled | montage — `hole_fill` cannot serve filled holes and deliberate gaps at once |
| position-dependent defocus | montage only |
| bright-field mask beyond ~10⁴ px | montage — the Fourier class will not fit in memory |

Both classes take the same kernels (`prlx`, `ssb`, `obf`, `mf`, `icom`) and agree numerically where they overlap. On the montage the convolution kernels are exact by default, evaluated by FFT on the canvas; `convolution_mode="stencil"` truncates them to a box instead. Truncating `icom` is the exception that is not an approximation — it gives **riCOM**, where the radius is a deliberate high-pass cutoff.

## What's new

- **`DirectPtychographyMontage`** with `from_dataset4d` / `from_virtual_bfs` / `from_dataset3d`, all five kernels, wrap and pad boundaries, and canvas upsampling that is bit-identical to constructing at a finer `scan_sampling`.
- **`FourierProbe`** — one interface over an analytic aperture-plus-aberrations probe and an empirical complex array. Makes a zone plate, a central stop, or a speckled X-ray optic usable as ψ(k). Sampled exactly by real-space zero-padding where the canvas and detector grids are commensurate, with bilinear interpolation as an opt-in fallback.
- **`wavelength=`** alongside `energy=`, since the electron de Broglie formula returns 0.137 Å for a 7.9 keV photon against the correct 1.569 Å.
- **`DirectPtychography.from_dataset3d`** — ungridded scans by regridding, with diagnostics and warnings for the two ways that goes wrong.
- **Position-dependent defocus** — `defocus_gradient`, `fit_defocus_gradient`, `defocus_map`.
- **Canvas pinning** (`obj_origin` / `obj_fov`) and `estimate_frame_drift`, so the frames of a multi-frame acquisition reconstruct onto pixel-identical canvases and can be aligned.
- **`rms_gradient_loss`** and a `loss=` parameter on both hyperparameter searches. Far better conditioned than the variance loss (28% dynamic range against 0.08% over a defocus series) and defined for every kernel, where `variance_loss` is parallax-only.
- **Performance** — grouped `conv2d` for stencils (14× over the tap loop), FFT convolution for kernels that span the canvas, and a detector collapse for riCOM that turns `num_bf` convolutions into two (14×/81×/397× at upsampling 1/2/4 on real data).

## Fixes to existing behavior

These change results for code already on `dev`:

- **`serialize`** recorded `type(value).__name__` as the container type, so a `torch.Size` was written as `"Size"` and rejected on load. `DirectPtychography.gpts` is `bf_mask.shape[:2]`, so **every reconstruction was unloadable**.
- **`dft_upsample` / `cross_correlation_shift`** applied the sub-pixel offset to the input frequency index rather than re-centering the output grid, carried the forward transform's sign, and converted the peak index against the wrong center. Worst-case error over shifts of 0–11.5 px falls from **2.0 px to 0.0022 px**. This also affects `DriftCorrection.align_translation` / `align_affine`.
- **`visualize`'s scalebar** ignored `upsampling_factor`, so an upsampled object reported double its true field of view.
- **`from_virtual_bfs`** defaulted `semiangle_cutoff=None`, which always raised a `TypeError` in the setter.

## Testing

`DirectPtychography` had no tests before this branch. Adds ~4,300 lines across `test_direct_ptychography.py`, `test_direct_ptychography_montage.py`, `test_imaging_utils.py` and a shared `conftest.py`.

370 passed / 26 skipped in `tests/diffractive_imaging`; 795 / 34 for the full suite.

Equivalence between the two classes is pinned throughout — the montage reproduces `DirectPtychography` to float precision on a gridded scan for every kernel, and the refactor in `b0dff8fc` was verified bit-identical across 35 outputs by running both trees in parallel `git worktree`s.

## References

- Shadow-montage (parallax) construction — *Microscopy and Microanalysis* **32**(1),
  ozaf126 (2026). https://doi.org/10.1093/mam/ozaf126
- riCOM, the truncated iCoM stencil — Yu et al., *Microscopy and Microanalysis* **28**,
  1526 (2022).

Related, though neither is the kernel implemented here — the first convolves a WDD kernel where this convolves SSB/OBF/MF kernels, and the second uses a segmented detector rather than a pixelated one:

- Convolution WDD — *Ultramicroscopy* **285**, 114411 (2026) https://doi.org/10.1016/j.ultramic.2026.114411
- Segmented-detector OBF — *Ultramicroscopy* **220**, 113133 (2021) https://doi.org/10.1016/j.ultramic.2020.113133

## Notes for review

- Tutorial notebooks are not included; they'll follow in `quantem-tutorials`.
- `ShadowMontagePtychography` was renamed to `DirectPtychographyMontage` before merge. It was introduced on this branch and never released, so nothing published depends on the old name.


arthurmccray:
It overall looks good to me! Only thing I would ask before merging into dev is that you update the tutorials (or at least make sure these changes don't break anything)

Also maybe clean up docstrings and comments to remove claudisms. e.g. https://github.com/electronmicroscopy/quantem/blob/921a1001619f006427d09a3f8be16dbfdb5f147b/src/quantem/diffractive_imaging/direct_ptychography.py#L284

gvarnavi:
Thanks @arthurmccray! Updated docstrings and tutorials.

New `Montage` tutorial here: https://github.com/electronmicroscopy/quantem-tutorials/blob/main/tutorials/diffractive_imaging/direct_ptycho_ungridded_scans.ipynb

arthurmccray (approved):
lgtm! thanks :)

PR #290, opened 2026-09-16
Add quantem.atoms: 3D atomic model analysis and atom tracing from volumes

cophus:

quantem had no tools for analyzing 3D atomic models, the output of atomic electron tomography (AET) or of atomistic simulations, and the atomic tracing on this branch stopped at finding peaks in a volume. This PR adds a `quantem.atoms` subpackage that takes a set of 3D coordinates through calibration, local structure classification, grain segmentation, strain, and visualization, and completes the tracer in `quantem.tomography.atom_trace` so that a reconstruction can be converted into such a model. No new required dependencies are added.

**Atom tracing (`quantem.tomography.Atoms`).** A volume is modeled as a sum of isotropic 3D Gaussians plus a smooth, non-negative Bezier background, and all site positions, amplitudes, widths and the background are optimized jointly with Adam against the volume. Each Gaussian is splatted only into a window of 3 to 4 sigma around its center, so the cost scales with the number of atoms rather than with the number of voxels. Sites are seeded from a difference-of-Gaussians filter and, during refinement, added at residual peaks, removed when weak or isolated, and merged when closer than a fraction of the nearest-neighbor spacing. On a 280^3 volume with 20,000 sites, 25 iterations take 4 s on an Apple GPU and 9 s on the CPU.

**Atomic model analysis (`quantem.atoms`).**
- `AtomicModel` stores the sites in a `Vector` with named per-site channels, so every measurement (scores, grain labels, strain components, intensities) is a column that the plots and the viewer can select. Coordinates from any file format are passed in as an `(N, 3)` array; the class has no format-specific loaders.
- `compute_pdf` and `calibrate` measure the radial distribution function with a KD-tree pair count, fit the first peak with an asymmetric generalized Gaussian, and set the physical voxel size from the nearest-neighbor distance of a reference crystal. `merge_close_sites` removes near-duplicate sites from tracing.
- `templates` generates polyhedral templates (fcc, hcp, bcc, sc, diamond/zincblende, wurtzite, icosahedral) from crystal definitions for any number of neighbor shells, together with their proper symmetry rotations, which are found numerically.
- `matching` classifies every site by polyhedral template matching, vectorized in torch. Trial rotations are built from pairs of neighbor vectors, each is scored by the one-to-one match between the rotated template and the measured neighbors, and the best rotation is refined with a batched Kabsch solve; an affine fit gives the deformation gradient. Both fcc and hcp templates take about 1 s for 20,000 sites on a laptop CPU. On synthetic fcc, hcp, bcc, diamond and wurtzite particles the classification accuracy is above 97% for interior sites at 3% bond-length noise and about 99% at 5%.
- `segment_grains` connects neighboring sites whose disorientation, reduced by the template symmetry, is below a threshold and labels the connected components, then fills the gaps between grains by majority vote; it also produces a `boundary` channel (interior, grain boundary, twin, surface). Twin boundaries in fcc appear as hcp-classified sites, since a site on a coherent {111} twin plane has the hcp environment.
- `fit_icosahedral_centers` locates the centers of multiply twinned particles as the least-squares intersection of the twin planes carried by the hcp sites, and `layer_positions` finds the atomic layers along any direction so that `plot("slices")` can step through a model one layer at a time.
- `compute_strain`, `surface_distance`, `sample_volume` and `classify_species` give per-site strain from the affine fit, the distance to the convex hull, the reconstruction intensity around each site, and a species assignment from a one-dimensional Gaussian mixture that leaves low-posterior sites unassigned.
- `structures` builds ideal Mackay icosahedra, cuboctahedra, Ino decahedra and mirror-twinned double icosahedra with a chosen center separation, with `shell` and `sector` channels, for comparison with measured models and for tests.
- `visualization` provides `model.plot(kind=...)` with the RDF, channel histograms, shaded-sphere slab projections, layer-indexed slice grids, and a template overlay for one site. `explode_grains` returns a copy with every grain displaced away from its center and the shared boundary sites duplicated into each adjacent grain.
- `model.show()` opens an interactive 3D viewer with channel selection, a histogram-driven color range, legend toggles, slab clipping along any normal with a draggable handle, a growth slider, and a site inspector. The viewer itself lives in the companion quantem.widget repository (separate PR); `quantem.atoms.show_atoms` only re-exports it.

`tests/atoms/test_atoms.py` adds 28 tests covering the templates (shell counts and symmetry orders), the RDF fit, the matcher on synthetic lattices (classification accuracy and rotation recovery), the measurements, the ideal structures, the center fit and layer indexing on a synthetic twinned particle, grain explosion, species assignment, and the end-to-end pipeline on a synthetic fcc twin. The tracer has no automated tests yet; it was checked on a reconstruction against an independently traced model, recovering 84% of the reference sites with a 0.6-voxel rms offset.

Two small changes outside the new package: `Dataset3d.show` gains `same_scale` so that all slices share one intensity range, and `quantem/__init__.py` imports `atoms`.

Review the public API of `AtomicModel` and the module layout, and try the pipeline on a 3D model of your own: `from_array`, `compute_pdf`, `calibrate`, `find_neighbors`, `match_templates(["fcc", "hcp"])`, `segment_grains`, then `plot("slab", channel="structure")`. Two points I would like feedback on:

1. Whether `quantem.atoms` should be a top-level subpackage, as here, or live under `quantem.tomography` next to the tracer. I put it at the top level because the analysis applies to any 3D atomic model, including simulation output.
2. The default classification rule. The soft score is the default; an RMSD-based rule and a neighbor-smoothed rule are available, and none of them is reliable above about 8% bond-length noise.

Open items before merging: the empty `tomography/atom_analysis.py` should be deleted, `atom_trace.py` needs tests, and the viewer PR in quantem.widget should land first so that `model.show()` resolves.

    - [ ] Documentation (e.g., tutorials, examples, README) has been updated.
    - [x] A tracking issue or plan to update documentation exists: a tutorial notebook on a synthetic twinned particle will follow once the API settles.

### DIFF
--- a/src/quantem/__init__.py
+++ b/src/quantem/__init__.py
@@ -10,5 +10,6 @@ from quantem.core import visualization as visualization
 
 from quantem import imaging as imaging
 from quantem import diffractive_imaging as diffractive_imaging
+from quantem import tomography as tomography
 
 __version__ = version("quantem")
--- a/src/quantem/core/datastructures/dataset3d.py
+++ b/src/quantem/core/datastructures/dataset3d.py
@@ -187,12 +187,13 @@ class Dataset3d(Dataset):
         start: int = 0,
         end: int | None = None,
         step: int = 1,
-        max: int | None = 20,
+        max_indices: int | None = 20,
         ncols: int = 4,
         scalebar: ScalebarConfig | bool = False,
         title_prefix: str | None = None,
         suptitle: str | None = None,
         returnfig: bool = False,
+        same_scale: bool = True,
         **kwargs,
     ) -> tuple[Figure, Axes] | None:
         """
@@ -203,10 +204,10 @@ class Dataset3d(Dataset):
         start : int, default 0
             First frame index. Supports negative indexing.
         end : int or None, optional
-            End frame index (exclusive). If None, determined by max.
+            End frame index (exclusive). If None, determined by max_indices.
         step : int, default 1
             Step between frames. Negative step shows frames in reverse order.
-        max : int or None, default 20
+        max_indices : int or None, default 20
             Maximum number of frames to show. Prevents memory issues.
             Set to None to show all frames.
         ncols : int, default 4
@@ -220,6 +221,10 @@ class Dataset3d(Dataset):
             Figure super title displayed above all subplots.
         returnfig : bool, default False
             If True, returns (fig, axes).
+        same_scale : bool, default True
+            If True, all slices share one intensity range (the volume's global
+            min/max) so they are directly comparable.  Ignored when you pass
+            ``vmin``/``vmax``/``norm``.  Set False for per-slice auto-contrast.
         **kwargs : dict
             Keyword arguments for show_2d (cmap, cbar, vmin, vmax, norm, etc.).
 
@@ -232,7 +237,7 @@ class Dataset3d(Dataset):
         Raises
         ------
         ValueError
-            If step is zero, ncols < 1, max < 1, start is out of bounds,
+            If step is zero, ncols < 1, max_indices < 1, start is out of bounds,
             or the specified range has no frames to display.
 
         Examples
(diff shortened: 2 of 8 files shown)


arthurmccray:
how the heck do i cancel a copilot review? i misclicked and now cant stop it 😢

PR #291, opened 2026-09-17
Update CONTRIBUTORS.md

arthurmccray:
Removed Colin Ophus from the Spectroscopy and Matthew Feikert from core.


arthurmccray:
only reason I didn't add @nwvlahakis was cuz nothing has made it to dev yet, which was the policy we agreed on when we first created this. Happy to make an exception in this case though cuz it'll be PRd in soon

